PR #40, opened 2019-09-19
fix: add CT target

Ulexus:
Cilium now [requires](https://github.com/cilium/cilium/pull/6862) NOTRACK.
The direct NOTRACK module is deprecated, and has been replaced by the CT
target, as mentioned
[here](https://github.com/cilium/cilium/issues/7435).

This enables the CT target.

Signed-off-by: Seán C McCord <ulexus@gmail.com>


andrewrynhard:
@Ulexus is this all that you needed? If you could do us one favor and update the drone config? We've' been updating as we go along. The channel field needs to be updated to get the build to pass:

`channel: proj-talos-maintainers`

Ulexus:
Yes, that is all I needed; Cilium works fine again, with this change.

PR #47, opened 2019-10-11
feat: add support for KVM to the kernel

andrewrynhard:


kernel/pkg.yaml
```diff
@@ -2,80 +2,80 @@ name: kernel
 variant: scratch
 shell: /toolchain/bin/bash
 dependencies:
-- image: docker.io/autonomy/tools:b4e3778
-- stage: ca-certificates
+    - image: docker.io/autonomy/tools:b4e3778
```
andrewrynhard: this is my YAML auto linter/formatter.

PR #52, opened 2019-10-12
fix: use util-linux stage for shared object dependency

andrewrynhard:
There are a few places where we are copying shared objects from the
toolchain's /lib. This moves to using the util-linux stage instead. This
ensures that all packages are being built against the version of
util-linux in this repo.

### DIFF
--- a/syslinux/pkg.yaml
+++ b/syslinux/pkg.yaml
@@ -2,43 +2,40 @@ name: syslinux
 variant: scratch
 shell: /toolchain/bin/bash
 dependencies:
-- stage: base
-- stage: musl
-- stage: ca-certificates
+  - stage: base
+  - stage: musl
+  - stage: ca-certificates
+  - stage: util-linux
 steps:
-- sources:
-  - url: https://www.kernel.org/pub/linux/utils/boot/syslinux/syslinux-6.03.tar.xz
-    destination: syslinux.tar.xz
-    sha256: 26d3986d2bea109d5dc0e4f8c4822a459276cf021125e8c9f23c3cca5d8c850e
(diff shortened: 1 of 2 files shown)


syslinux/pkg.yaml
```diff
@@ -2,43 +2,40 @@ name: syslinux
...
+        tar -xJf syslinux.tar.xz --strip-components=1
 
-    mkdir /bin
-    ln -sv /toolchain/bin/bash /bin/bash
-    ln -sv /toolchain/bin/bash /bin/sh
-    ln -sv /toolchain/bin/pwd /bin/pwd
-    cp -R /toolchain/lib/gcc /lib
-    cp -R /toolchain/lib/libgcc* /lib
-
-    # TODO(andrewrynhard): Add a util-linux package in this repo and pull these from there.
```
andrewrynhard: Removed.

xfsprogs/pkg.yaml
```diff
@@ -2,38 +2,34 @@ name: xfsprogs
...
-    ln -sv /toolchain/bin/bash /bin/bash
-    ln -sv /toolchain/bin/bash /bin/sh
-    cp -R /toolchain/lib/gcc /lib
-    cp -R /toolchain/lib/libgcc* /lib
+        ln -sv /toolchain/bin/bash /bin/bash
+        ln -sv /toolchain/bin/bash /bin/sh
+        cp -R /toolchain/lib/gcc /lib
+        cp -R /toolchain/lib/libgcc* /lib
 
-    # TODO(andrewrynhard): Add a util-linux package in this repo and pull these from there.
```
andrewrynhard: Removed.

PR #64, opened 2019-10-14
chore: cleanup ca-certificates dependency

smira:
It's not actually a dependency for the build, as downloads are performed
via buildkit, not as part of the build itself.


andrewrynhard:
Nice.

PR #65, opened 2019-10-14
refactor: add 'musl' as runtime dependency for 'base'

smira:
This dries up dependency tree in `pkg.yaml` and graphs.

See https://github.com/talos-systems/bldr/pull/33


smira:
WIP until bldr PR is merged and shebang can be updated

base/pkg.yaml
```diff
@@ -4,6 +4,7 @@ shell: /toolchain/bin/bash
 dependencies:
   - image: docker.io/autonomy/tools:7de4db2
   - stage: musl
+    runtime: yes
```
andrewrynhard: Cool, so this says "anytime `base` is a dependency, pull `musl` in too"?

smira: yep, exactly, and it works transitively and deduplicates stages as required.

we can actually pull this `tools` dependency as runtime and avoid putting it in every stage

I'm torn whether it's nice or not

smira: actually it's already pulled similar way by `finalize` which does copy `/` -> `/`, so `base` contains `tools`

PR #66, opened 2019-10-14
chore: use `Pkgfile` vars to expose base image better

smira:
See talos-systems/bldr#34


smira:
Draft due to the shebang

PR #74, opened 2019-11-27
feat: upgrade ca-certificates, eudev, libressl, and libseccomp

andrewrynhard:


smira:
looks like ca-certificates got updated and they have new hash

andrewrynhard:
> looks like ca-certificates got updated and they have new hash

Thanks for that catch!

PR #89, opened 2020-01-20
feat: Add tzdata package

bradbeam:
Generate Etc/UTC timezone

Signed-off-by: Brad Beam <brad.beam@talos-systems.com>


bradbeam:
closes https://github.com/talos-systems/talos/issues/1477

bradbeam:
guess its not worth the effort. 
go will default to UTC which covers most of our bases. 
ref https://golang.org/src/time/zoneinfo_unix.go

PR #90, opened 2020-01-23
feat: enable CONFIG_BRIDGE_VLAN_FILTERING

andrewrynhard:
This enables VLAN filtering.


kernel/pkg.yaml
```diff
@@ -2,90 +2,90 @@ name: kernel
 variant: scratch
 shell: /toolchain/bin/bash
 dependencies:
-    - image: '{{ .TOOLS_IMAGE }}'
```
andrewrynhard: All the changes here are from a YAML auto formatter.

PR #94, opened 2020-01-27
feat: add i40e driver support

andrewrynhard:
This adds support for Intel i40e ethernet drivers.


andrewrynhard:
Decided to cherry-pick instead.

PR #108, opened 2020-03-24
feat: enable USB storage support

rsmitty:
This PR will allow for users to attach USB devices to Talos machines and
interact with them.


kernel/config-amd64
```diff
@@ -3932,7 +3932,22 @@ CONFIG_USB_EHCI_PCI=y
...
+# CONFIG_USB_STORAGE_USBAT is not set
+# CONFIG_USB_STORAGE_SDDR09 is not set
+# CONFIG_USB_STORAGE_SDDR55 is not set
+# CONFIG_USB_STORAGE_JUMPSHOT is not set
+# CONFIG_USB_STORAGE_ALAUDA is not set
+# CONFIG_USB_STORAGE_ONETOUCH is not set
+# CONFIG_USB_STORAGE_KARMA is not set
+# CONFIG_USB_STORAGE_CYPRESS_ATACB is not set
+# CONFIG_USB_STORAGE_ENE_UB6250 is not set
+# CONFIG_USB_UAS is not set
```
andrewrynhard: Do we need to enable these?

rsmitty: they aren't enabled, the only one that actually gets a `y` flag is CONFIG_USB_STORAGE. the others just remain in `not set`.

andrewrynhard: RIght, but I guess I don't know if we need to enable these and `s/is not set/y`.

PR #126, opened 2020-05-29
feat: build arm64 packages

andrewrynhard:
This updates builds for arm64.


smira:
Merged kernel config into #142

PR #129, opened 2020-06-12
fix: publish libaio and disable buffering

andrewrynhard:
LVM2 depends on libaio. This patch ensures that we publish libaio.
There also is a change in newer versions of musl that causes LVM2
to segfault because stdout/stderr is closed. This adds a patch that
disables buffering in create_toolcontext function to prevent the
segfault.


andrewrynhard:
I found https://github.com/osresearch/heads/pull/652 as a reference for this PR.

PR #143, opened 2020-07-21
feat: add grub

andrewrynhard:


grub/pkg.yaml
```diff
@@ -0,0 +1,32 @@
...
+                    --sysconfdir=/etc \
+                    --disable-efiemu \
+                    --disable-werror
+    build:
+      - |
+        make -j $(nproc)
+    install:
+      - |
+        make install DESTDIR=/rootfs
+        mkdir -p /rootfs/usr/share/bash-completion/completions
```
smira, line 28: do we really need bash completion?

andrewrynhard: Not in Talos. But this package should be able to stand on its' own. I envision that other distros might be able to help us maintain these packages in the future. Moving them to `/usr/share` ensures that they get removed by the Talos build: https://github.com/talos-systems/talos/blob/master/hack/cleanup.sh#L31.

PR #145, opened 2020-07-27
feat: add FUSE kernel support

rsmitty:
This PR enables FUSE support at the kernel level. Userspace libraries
are still needed for it to work correctly, but that burden is on the
user.


smira:
I guess we need to change pipeline type to `kubernetes` now...

PR #146, opened 2020-08-04
fix: update arm64 kernel config with proper source

smira:
This uses our toolchain/kernel version to drive the config.

The most important change was to enable ARM serial console (AMBA
driver).


smira:
I can try to go and take out any options which seem to be not relevant, as I might have enabled a bit more than we need.

andrewrynhard:
> I can try to go and take out any options which seem to be not relevant, as I might have enabled a bit more than we need.

That might be a good idea. One thing we need to figure out is how to have common config options that we need for x86 and ARM. If we wanted to start like that, what I did was I ran `make olddefconfig ARCH=aarch64` (or something like that) and it used the x86 config as a base, and output an ARM config. We should probably do that, and only add the serial console option specific to ARM. That way we know we have the two configs more or less aligned.

smira:
> > I can try to go and take out any options which seem to be not relevant, as I might have enabled a bit more than we need.
> 
> That might be a good idea. One thing we need to figure out is how to have common config options that we need for x86 and ARM. If we wanted to start like that, what I did was I ran `make olddefconfig ARCH=aarch64` (or something like that) and it used the x86 config as a base, and output an ARM config. We should probably do that, and only add the serial console option specific to ARM. That way we know we have the two configs more or less aligned.

that was I aсtually did, but I see that eventually we'll have more and more options custom for ARM - support for different kinds of hardware: Raspberry PIs, etc. 

when we upgrade kernel verison, if we always do (and automate)  `make olddefconfig`, it should take us forward in a good way from each config and it would also allow us to be up to date.

smira:
runc release was updated upstream due to LGPL issues, I'll create a PR to fix the build: https://github.com/opencontainers/runc/releases/tag/v1.0.0-rc90

smira:
I'll rebuild this with #148

PR #159, opened 2020-09-03
fix: build `hwdb.bin` from hwdb files in eudev package

smira:
See e.g. http://www.linuxfromscratch.org/lfs/view/9.1/chapter06/eudev.html

Signed-off-by: Andrey Smirnov <smirnov.andrey@gmail.com>

PR #160, opened 2020-09-03
fix: disable hwdb in eudev

smira:
Hardware DB might make little sense for servers, so disable it to
reduce initramfs size.


smira:
Competitor to #159

PR #162, opened 2020-09-04
feat: upgrade containerd to v1.4.0

andrewrynhard:
This brings in the latest stable containerd.


smira (approved):
Do we have to enable anything for cgroups v2?

andrewrynhard:
> Do we have to enable anything for cgroups v2?

Not that I am aware of. Cgroups2, I think still needs to be implemented in the kubelet. So we may be waiting for a while to use it.

PR #184, opened 2020-11-07
feat: add R8169 driver

andrewrynhard:
Adds Realtek 8169/8168/8101/8125 ethernet support.


Ulexus (approved):

Ulexus:
My thought is that this chip is vastly less-used on ARM, but there is no real reason to make the configs more different than they have to be. :+1:

andrewrynhard:
> My thought is that this chip is vastly less-used on ARM, but there is no real reason to make the configs more different than they have to be.

Oh totally agree. Yeah the strategy thusfar has been to try to keep them in sync as much as we can.

PR #189, opened 2020-11-19
feat: add u-boot

andrewrynhard:
This adds u-boot.

### DIFF
--- a/Pkgfile
+++ b/Pkgfile
@@ -3,7 +3,7 @@
 format: v1alpha2
 
 vars:
-  TOOLS_IMAGE: ghcr.io/talos-systems/tools:v0.3.0-10-g83dc352
+  TOOLS_IMAGE: ghcr.io/talos-systems/tools:v0.3.0-11-g84c834e
 
 labels:
   org.opencontainers.image.source: https://github.com/talos-systems/pkgs
--- a/u-boot/pkg.yaml
+++ b/u-boot/pkg.yaml
@@ -1,0 +1,102 @@
+# References:
+#   U-Boot:
+#     - https://u-boot.readthedocs.io/en/latest
+#   Allwinner:
+#     - https://linux-sunxi.org/
+#     - https://github.com/u-boot/u-boot/blob/master/board/sunxi/README.sunxi64
+#   Rockchip:
+#     - http://opensource.rock-chips.com
+#     - https://github.com/u-boot/u-boot/blob/master/doc/board/rockchip/rockchip.rst
+name: u-boot
+variant: scratch
+shell: /toolchain/bin/bash
+dependencies:
+  - stage: base
+steps:
+  - sources:
+      - url: https://github.com/ARM-software/arm-trusted-firmware/archive/v2.4.tar.gz
+        destination: arm-trusted-firmware.tar.gz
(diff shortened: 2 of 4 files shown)


Pkgfile
```diff
@@ -3,7 +3,7 @@
 format: v1alpha2
 
 vars:
-  TOOLS_IMAGE: ghcr.io/talos-systems/tools:v0.3.0-10-g83dc352
+  TOOLS_IMAGE: docker.io/andrewrynhard/tools:v0.3.0-3-g6b491a6-dirty
```
andrewrynhard, line 6: I have a PR out for this.

u-boot/pkg.yaml
```diff
@@ -0,0 +1,26 @@
+name: u-boot
```
andrewrynhard, line 1: Since u-boot has different configurations for different boards, we might want to name this `u-boot-libretech-all-h3-cc-h5`. Alternative is that we build all the variations under `u-boot`.

smira: interesting if we could have built multiple variants and could choose one or another

smira (approved):

u-boot/pkg.yaml
Ulexus: Just to note:  there appear to be hundreds of variations.

andrewrynhard: yeah :/ ... not sure what the best strategy is.

Pkgfile
```diff
@@ -3,7 +3,7 @@
 format: v1alpha2
 
 vars:
-  TOOLS_IMAGE: ghcr.io/talos-systems/tools:v0.3.0-10-g83dc352
+  TOOLS_IMAGE: ghcr.io/talos-systems/tools:v0.3.0-11-g84c834e
```
andrewrynhard, line 6: Brings in python3.

u-boot/pkg.yaml
andrewrynhard: For now, I think we can do is just build variants here, and add all of them to the installer image. We can figure out something better in the long run.

Ulexus:
FWIW, the rpi3 and rpi4 have their own 64-bit versions, which are distinct from `rpi_arm64`.

andrewrynhard:
> FWIW, the rpi3 and rpi4 have their own 64-bit versions, which are distinct from `rpi_arm64`. The main difference is that the rpi4 version includes the driver for the emmc.

I wonder which we we actually need. Do you happen to know? That SO answer suggests `rpi_arm64`.

Ulexus:
I suspect both will generally work for most cases.  The rpi_arm64 one seems to include most everything which rpi_4 has except for eMMC (and maybe SD... can't really tell for sure), while rpi_arm64 has the older USB-attached ethernet which previous generations of the Pi used.

My off-the-cuff thought is that we should use the platform-specific ones.

andrewrynhard:
> I suspect both will generally work for most cases. The rpi_arm64 one seems to include most everything which rpi_4 has except for eMMC (and maybe SD... can't really tell for sure), while rpi_arm64 has the older USB-attached ethernet which previous generations of the Pi used.
> 
> My off-the-cuff thought is that we should use the platform-specific ones.

:+1: sounds good

smira (approved):

PR #195, opened 2020-11-24
feat: Enable NF_CONNTRACK_TFTP

e-nomem:
I'm hoping to use the Sidero TFTP service with metallb/BGP without having to use host networking for the pod, but TFTP is a fun protocol and requires a helper to handle connection tracking properly. Enabling this kernel option/module should be sufficient, but I haven't had a chance to test it yet.


smira:
@e-nomem looks like I can't push to your branch with fixes, so I will open new PR

smira:
Updated kernel configs in #196 via `make kernel-olddefconfig PLATFORM=linux/amd64,linux/arm64`

andrewrynhard:
@e-nomem Let me know if this works, I have been wanting to do the same.

e-nomem:
I thought I had 'Allow edits by maintainers' allowed... In any case, I'll keep you posted on my testing.

PR #210, opened 2020-12-03
regulators

andrewrynhard:
- feat: enable more power regulators
- test


kernel/kernel/config-arm64
```diff
@@ -392,7 +397,7 @@ CONFIG_ARCH_ENABLE_SPLIT_PMD_PTLOCK=y
 CONFIG_SECCOMP=y
 CONFIG_PARAVIRT=y
 # CONFIG_PARAVIRT_TIME_ACCOUNTING is not set
-# CONFIG_KEXEC is not set
+CONFIG_KEXEC=y
```
smira, line 400: do we really want to enable it?

kernel/kernel/config-arm64
```diff
@@ -724,10 +741,11 @@ CONFIG_CRYPTO_AES_ARM64_BS=y
 #
 # General architecture-dependent options
 #
+CONFIG_CRASH_CORE=y
```
smira, line 744: same question here

kernel/kernel/config-arm64
```diff
@@ -883,58 +903,9 @@ CONFIG_MQ_IOSCHED_KYBER=y
 # end of IO Schedulers
 
 CONFIG_PREEMPT_NOTIFIERS=y
+CONFIG_PADATA=y
 CONFIG_ASN1=y
-CONFIG_ARCH_INLINE_SPIN_TRYLOCK=y
```
smira, line 887: not sure if we need to remove those inlines?

kernel/kernel/config-arm64
andrewrynhard: No, I got a list of everything added and removed when I ran `make olddefconfig`. I plan on going through that today and fixing up things we don't want to enable.

PR #214, opened 2020-12-09
feat: add KSPP compliance check script

smira:
We tailor results for our needs. I picked Python for the script as we
don't have jq in the tools ;)

Also fixed the kernel config to be compliant according to the script and
subset of the enforcements.

I had to add a patch to kernel's Kconfig files to switch off some
default options which otherwise get enabled via `olddefconfig` and
violate KSPP.


andrewrynhard (approved):

smira:
just to clarify: if our config violates KSPP (as defined by the script), build fails now

PR #222, opened 2020-12-17
chore: drop vmlinux from kernel package

smira:
We're not using uncompressed kernel anywhere, it's huge and it can't be
used directly in most of the cases.


Unix4ever (approved):
LGTM, but probably title should say `drop vmlinux` :)

PR #227, opened 2020-12-28
feat: enable CONFIG_IPV6_MULTIPLE_TABLES option

smira:
Wireguard depends on it.


Unix4ever (approved):
:+1: I hope that's the only flag we need to enable :)

PR #233, opened 2021-01-19
feat: enable hyperv_utils in linux kernel

willemm:
CONFIG_HYPERV_UTILS=y in the amd64 kernel config

Enabling this in the kernel makes running it on Hyper-V work better:

It enables running a hyper-v integration daemon, which is needed to enable things like setting a static IP address, copying files onto the server from the virtual machine manager.

To my surprise, it also means the wallclock time gets set correctly (as opposed to shifted by an hour because we're in GMT+1), so I guess it enables automatic time sync to the hyper-v host.

### DIFF
--- a/kernel/kernel/config-amd64
+++ b/kernel/kernel/config-amd64
@@ -4463,7 +4463,7 @@ CONFIG_VHOST_VSOCK=y
 #
 CONFIG_HYPERV=y
 CONFIG_HYPERV_TIMER=y
-# CONFIG_HYPERV_UTILS is not set
+CONFIG_HYPERV_UTILS=y
 # CONFIG_HYPERV_BALLOON is not set
 # end of Microsoft Hyper-V guest support
 


smira:
I'll fix up arm64 config

smira:
> I'll fix up arm64 config

it doesn't have to be changed, so all good :+1:

smira (approved):

smira:
@willemm would you mind re-pushing the PR to conform with conventional commits (https://www.conventionalcommits.org/en/v1.0.0/). I added a PR check for it, it should get triggered on next push

basically commit header should be `feat: enable hyperv_utils in Linux kernel`

and commit itself should have DCO, (`git commit --amend -s` should help you with that)

willemm:
Something new learned every day :)  I edited the commit message, and I think it's signed off correctly now?

smira:
> Something new learned every day :) I edited the commit message, and I think it's signed off correctly now? Or did I now completely mess it up by force-pushing

thanks @willemm, it looks good now!

PR #235, opened 2021-02-02
feat: add cryptsetup dependencies to all targets

Unix4ever:
To build them as separate containers and then install into Talos along
with cryptsetup.
Also enable couple kernel flags to fix `cryptsetup luksOpen`.

Signed-off-by: Artem Chernyshev <artem.0xD2@gmail.com>


Makefile
```diff
@@ -17,7 +17,7 @@ COMMON_ARGS += --platform=$(PLATFORM)
 empty :=
 space = $(empty) $(empty)
 
-TARGETS =  ca-certificates  cni  containerd cryptsetup dosfstools  eudev  fhs  grub ipmitool  iptables  kernel  kmod  libaio  libressl  libseccomp  linux-firmware lvm2  musl  open-iscsi  open-isns raspberrypi-firmware runc  socat  syslinux u-boot  util-linux  xfsprogs
+TARGETS =  ca-certificates  cni  containerd cryptsetup dosfstools  eudev  fhs  grub ipmitool  iptables  kernel  kmod  libaio  libressl  libseccomp  linux-firmware lvm2  musl  open-iscsi  open-isns raspberrypi-firmware runc  socat  syslinux u-boot  util-linux  xfsprogs  libpopt libjson-c
```
smira, line 20: I guess they should be better sorted

Unix4ever: rgr

Unix4ever: I guess I also need some changes in kernel flags. So I'm going to push it into that PR as well

smira (approved):

PR #238, opened 2021-02-06
feat: bump raspberrypi-firmware

AlekSi:


smira:
@AlekSi could you please amend commit with DCO and something in the body? 

(linter doesn't work, trying to figure out why)

PR #246, opened 2021-02-23
feat: upgrade kernel to 5.11.1

rsmitty:
This PR pulls in the latest stable kernel.

Signed-off-by: Spencer Smith <robertspencersmith@gmail.com>


andrewrynhard (changes requested):
Blocking the merge just in case we merge on accident. We have to wait until 0.9 is cut.

smira:
I would even keep 5.10 for 0.10, as it's going to be LTS branch

PR #249, opened 2021-03-04
feat: add u-boot for Rock Pi 4

Unix4ever:
This builds u-boot for the Rock Pi 4.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.bin filter=lfs diff=lfs merge=lfs -text
--- a/u-boot/pkg.yaml
+++ b/u-boot/pkg.yaml
@@ -26,10 +26,12 @@ steps:
     env:
       SUN50I_A64_ARM_TRUSTED_FIRMWARE: sun50i_a64_arm-trusted-firmware
       RK3328_ARM_TRUSTED_FIRMWARE: rk3328_arm-trusted-firmware
+      RK3399_ARM_TRUSTED_FIRMWARE: rk3399_arm-trusted-firmware
       LIBRETECH_ALL_H3_CC_H5_U_BOOT: libretech_all_h3_cc_h5_u-boot
       ROCK64_RK3328_U_BOOT: rock64_rk3328_u-boot
       RPI_ARM64_U_BOOT: rpi_4_u-boot
       BANANAPI_M64_U_BOOT: bananapi_m64_u-boot
+      ROCK_PI_4_RK3399_U_BOOT: rock-pi-4-rk3399_u-boot
     prepare:
       - |
         mkdir -p /usr/bin \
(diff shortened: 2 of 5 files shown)


Unix4ever:
So I just added prebuilt binaries for the part which does not build with the regular gcc and now it's all good.
These binaries are tiny anyway: it's much less hassle than adding gcc-arm-eabi-none

smira (commented):
do we need all these files for `m0` or just the final `bin` file is enough?

Unix4ever:
> do we need all these files for m0 or just the final bin file is enough?

I wasn't sure. Let me check

Unix4ever:
yeah, only bin is enough.

AlekSi:
GitHub shows two files in this PR as empty. Is it GitHub's bug?

AlekSi:
They were not empty in the previous commits, so I guess that's not GutHub's bug, but commit bug :)

Also, what do you think about using Git LFS for binary files? They are small, but if we ever change them, the total repo size will grow.

andrewrynhard:
> They were not empty in the previous commits, so I guess that's not GutHub's bug, but commit bug :)
> 
> Also, what do you think about using Git LFS for binary files? They are small, but if we ever change them, the total repo size will grow.

I have a similar concern. Also, we should document how to build these bin files in case we ever need to again. I am a bit torn about adding binary files TBH.

Unix4ever:
Yeah... Another option was to add them somewhere to releases somewhere.

Unix4ever:
> Also, what do you think about using Git LFS for binary files? They are small, but if we ever change them, the total repo size will grow.

hmm. That's weird. They don't show as empty on my local clone.
Will try LFS

u-boot/pkg.yaml
```diff
@@ -26,10 +26,13 @@ steps:
     env:
       SUN50I_A64_ARM_TRUSTED_FIRMWARE: sun50i_a64_arm-trusted-firmware
       RK3328_ARM_TRUSTED_FIRMWARE: rk3328_arm-trusted-firmware
+      RK3399_ARM_TRUSTED_FIRMWARE: rk3399_arm-trusted-firmware
       LIBRETECH_ALL_H3_CC_H5_U_BOOT: libretech_all_h3_cc_h5_u-boot
       ROCK64_RK3328_U_BOOT: rock64_rk3328_u-boot
       RPI_ARM64_U_BOOT: rpi_4_u-boot
       BANANAPI_M64_U_BOOT: bananapi_m64_u-boot
+      ROCK_PI_4_RK3399_U_BOOT: rock-pi-4-rk3399_u-boot
+      M0_CROSS_COMPILE: /toolchain/bin/
```
andrewrynhard, line 35: I don't see where this is used?

Unix4ever: well, it's used in u-boot makefiles but as we skip the build I guess I should drop it, thanks :+1:

.gitattributes
```diff
@@ -0,0 +1,2 @@
+u-boot/rockpi4/m0/rk3399m0.bin filter=lfs diff=lfs merge=lfs -text
+u-boot/rockpi4/m0/rk3399m0pmu.bin filter=lfs diff=lfs merge=lfs -text
```
AlekSi, line 2: I think that can be just `*.bin` if we agree to use `bin` extension for binary files only and always use Git LFS for them

Unix4ever: sure, updated

smira (approved):

PR #255, opened 2021-03-11
chore: do not use git-lfs for rockpi4 binaries

Unix4ever:
Looks like an overkill for such a small files, while it brings issues
in our CI pipeline as you have to install git-lfs there.

Signed-off-by: Artem Chernyshev <artem.0xD2@gmail.com>


smira:
restarted the build

smira (approved):

PR #265, opened 2021-04-15
feat: add cpu scaling for rpi

rsmitty:
This PR ensures we support scaling CPU on the rpi. Results in 3x the amount of CPU events processed each second!

Signed-off-by: Spencer Smith <robertspencersmith@gmail.com>


rsmitty:
Of other possible notes, we use the "schedutil" governor by default which seems to be the correct default for ARM nodes (https://www.phoronix.com/scan.php?page=news_item&px=Linux-5.9-Power-Management).

I've also noticed there's a file in raspbian called `/etc/init.d/raspi-config` which forces the governor to "ondemand" and sets some thresholds for that governor:

```
        echo 50 > /sys/devices/system/cpu/cpufreq/ondemand/up_threshold
        echo 100000 > /sys/devices/system/cpu/cpufreq/ondemand/sampling_rate
        echo 50 > /sys/devices/system/cpu/cpufreq/ondemand/sampling_down_factor

```

This appears to offer a small benefit, but nothing I'd consider worth doing right now (~1735 events per second vs. ~1739-1740 events per second).

Unix4ever (approved):

smira (approved):
we should also get this into 0.10 (backport to release-0.5 branch)

PR #267, opened 2021-04-27
feat: add iPXE build with embedded placeholder script

smira:
This builds iPXE with HTTPS enabled, and embedded placeholder script
which is to be replaced with the real script during Sidero
metal-controller-manager bootstrap.


ipxe/pkg.yaml
```diff
@@ -0,0 +1,47 @@
...
+        {{ if eq .ARCH "aarch64" }}
+        cp -p bin-arm64-efi/ipxe.efi /rootfs/usr/libexec
+        {{ else }}
+        cp -p bin/undionly.kpxe bin-x86_64-efi/ipxe.efi /rootfs/usr/libexec
+        cp -p bin/undionly.kpxe.bin bin/undionly.kpxe.zinfo util/zbin /rootfs/usr/libexec/zbin
+        {{ end }}
+finalize:
+  - from: /rootfs
+    to: /
+# Generating .kpxe via zbin:
```
Unix4ever, line 43: is that some leftovers?

Unix4ever (approved):

ipxe/pkg.yaml
smira: no, that is doc on how to generate `.kpxe`.. 

so the thing is that we want to patch binary in place, but `.kpxe` is lzma-compressed; so we generate uncompressed version, a tool to generate compressed version, and we patch uncompressed version and compress it on the fly; thsee are the steps to compress the uncompressed version to build final `.kpxe`

PR #278, opened 2021-05-14
Add geneve encapsulation support for openvswitch

sbskas:
geneve encapsultation is missing from the talos kernel.
This patch add support for it.
Tested on kube-ovn.


smira:
Could you please squash commits and use conventional commit format + DCO?

smira (approved):

PR #280, opened 2021-05-18
feat: add support for netxen card

sbskas:
This driver is mandatory to boot talos on DL580 G7.
No extra firmware required.


rsmitty:
@sbskas thanks for the PR. I'll merge this and let it build. It should be pulled in as part of another pkg update to upstream Talos later this afternoon. I'll CC you on that PR when it pops up.

PR #315, opened 2021-08-09
feat: add 'mdadm' package

sergelogvinov:
I am playing with soft raid as second data store and this package helps me with that.

udev rules activate healthy mdraid automatically. 

May be this is too early for talos, but I keep it here.


smira:
We have `mdadm` in extensions now, so I think this is obsolete.

PR #334, opened 2021-09-06
feat: enable KEXEC_FILE_LOAD in the kernel

smira:
We don't enable `KEXEC_LOAD`, but only enable kexecing from the file
image, which is available on amd64/arm64.


andrewrynhard (approved):
Here we go 🤞🏻

PR #348, opened 2021-11-04
feat: update cpu governor to schedutil

rsmitty:
This PR updates the default governor to "schedutil" away from
"userspace" on amd64 systems. Schedutil seems to be the default on
Ubuntu systems and the eventual default on all distros (already the
default on all ARM distros, including ours). I noticed no real
difference in sysbench testing between schedutil and ondemand, so I
believe this is a sane default choice for us. There was a very big diff
between the old value of userland, however, since there was no userland
process actually managing the frequencies.

Signed-off-by: Spencer Smith <spencer.smith@talos-systems.com>


rsmitty:
some interesting before/after notes:

```
## Sysbench command

sysbench --test=memory --memory-total-size=50G --memory-block-size=4M --time=10 run

## Ubuntu Live CD

- 51200.00 MiB transferred (11077.36 MiB/sec)

## Default Talos Install (privileged ubuntu pod)

- Default configured scaling governor is "userspace".
- 51200.00 MiB transferred (10717.62 MiB/sec)

## Tweaked Talos Install (privileged ubuntu pod)

- Override scaling governor to be "ondemand" with  `echo ondemand | tee /sys/devices/system/cpu/cpu*/cpufreq/scaling_governor`
- 51200.00 MiB transferred (11601.04 MiB/sec)
```

kpostrup:
I would like to add that this change will primarily be noticeable when running CPU benchmarks.

Using the 7z benchmark, e.g. `7z b -mmt1` (from the **p7zip-full**), which runs single-threaded, I've seen a performance increase of 350%.

Unix4ever (approved):

PR #356, opened 2021-11-16
feat: add mdraid 1/0/10

sergelogvinov:
It helps to use soft raid as second (data) drive.

Very useful for GCP local SSDs or Equinix Metal.

RAID 0/1/10 does not affect boot time.


smira:
Just curious, how is it supposed to be used with Talos?

sergelogvinov:
It helps csi to create/prepare/mount soft raid to the pod

As talos already has lvm kernel support.

smira:
So it's not about Talos system disk, but rather about additional disks and volume provisioners, got it.

smira (approved):

PR #363, opened 2021-12-01
feat: update CNI to 1.0.1, separate package for flannel

smira:
Flannel got moved out to its own repository, so we have to build it as a
separate package.


smira:
there will be a follow-up PR to `extras` to rebuild the bundle with `flannel`

rsmitty (approved):

PR #374, opened 2021-12-26
feat: (kernel) enable amdgpu

anthr76:
Enable amdgpu in the kernel in order to support https://github.com/RadeonOpenCompute/k8s-device-plugin

Considering i915 support is present there should be no reason why AMDGPU isn't enabled?

Signed-off-by: Anthony Rabbito <hello@anthonyrabbito.com>


smira (approved):
thank you!

PR #386, opened 2022-01-27
chore: bump kernel to 5.15.17

frezbo:
Bump kernel to 5.15.17


smira:
we run `make kernel-olddefconfig` to make sure config is updated to the new kernel version, sometimes there are small updates

smira (approved):

PR #388, opened 2022-01-31
chore: bump kernel to 5.15.18

frezbo:
Bump kernel to 5.15.18 stable


smira:
we still need `make kernel-olddefconfig` :)

PR #393, opened 2022-02-09
feat: add Intel Ethernet Controller I225-V driver

smira:
https://linux-hardware.org/?id=pci:8086-15f3-1043-87d2


Ulexus (approved):

kernel/build/config-arm64
```diff
@@ -2784,7 +2784,7 @@ CONFIG_IAVF=y
 CONFIG_I40EVF=y
 # CONFIG_ICE is not set
 # CONFIG_FM10K is not set
-# CONFIG_IGC is not set
+CONFIG_IGC=y
```
frezbo, line 2787: do we need these NUC drivers for ARM?

frezbo (approved):

kernel/build/config-arm64
smira: we tried to keep configs in sync across arches, if not something which can't be on a specific platform

I don't know if there's any ARM computer with that network card, but that's possible I guess?

frezbo: yeh, makes sense to keep both in sync

PR #394, opened 2022-02-10
chore: disable TIPC and B.A.T.M.A.N

frezbo:
TIPC has been known to vulnerabilities in the past,
- [CVE-2021-43267](https://nvd.nist.gov/vuln/detail/CVE-2021-43267)
- [CVE-2022-0435](https://github.com/torvalds/linux/commit/9aa422ad326634b76309e8ff342c246800621216)

TIPC also requires the `tipc` kernel module to be enabled. It does not
make sense to ship it with kernel

Also disabled [B.A.T.M.A.N](https://www.open-mesh.org/projects/batman-adv/wiki/Wiki) since it's
mostly used for wireless mesh

Signed-off-by: Noel Georgi <git@frezbo.dev>

PR #397, opened 2022-02-11
chore: bump kernel to 5.15.23

frezbo:
Bump kernel to 5.15.23

Fixes CVE-2022-0435, as of
https://github.com/talos-systems/pkgs/pull/394 Talos is not affected

### DIFF
--- a/kernel/build/config-amd64
+++ b/kernel/build/config-amd64
@@ -1,6 +1,6 @@
 #
 # Automatically generated file; DO NOT EDIT.
-# Linux/x86 5.15.22 Kernel Configuration
+# Linux/x86 5.15.23 Kernel Configuration
 #
 CONFIG_CC_VERSION_TEXT="gcc (GCC) 11.2.0"
 CONFIG_CC_IS_GCC=y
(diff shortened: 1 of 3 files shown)


smira (approved):

smira:
I guess we need to backport kernel bump to release-0.9 as well. Probably not the #394 to avoid big (breaking?) changes

frezbo:
I'm not sure, we probably don't have anything using TIPC in userspace, but safer to update

frezbo:
> I guess we need to backport kernel bump to release-0.9 as well. Probably not the #394 to avoid big (breaking?) changes

https://www.openwall.com/lists/oss-security/2022/02/10/1 since we don't load it and have KASLR enabled maybe we can skip backporting

smira:
> It's also worth noting that the current `CONFIG_FORTIFY_SRC=y` is a hard
mitigation to leveraging CVE-2022-0435 for control-flow hijacking, as it
does a bounds check on the size of the offending memcpy and causes a
kernel panic.

I think loading is not relevant, as we have built into the kernel, but on other hand probably yes, it's only panic potentially. But at the same time it might make to bump to address issues.

smira:
I will create backport for 5.15.23 to release-0.9

PR #402, opened 2022-02-17
feat: add openssl

andrewrynhard:


smira:
I'll take over this PR to replace LibreSSL with OpenSSL in pkgs.

Makefile
```diff
@@ -19,7 +19,40 @@ COMMON_ARGS += --build-arg=https_proxy=$(https_proxy)
 empty :=
 space = $(empty) $(empty)
 
-TARGETS =  ca-certificates  cni  containerd cryptsetup dosfstools  eudev  fhs flannel-cni grub libinih liburcu ipmitool  iptables ipxe kernel  kmod  libaio libjson-c liblzma libpopt libressl  libseccomp  linux-firmware lvm2  musl  open-iscsi  open-isns raspberrypi-firmware runc  socat  syslinux u-boot  util-linux  xfsprogs
+TARGETS = \
+	ca-certificates \
```
Unix4ever, line 23: I don't see `openssl` in the list, should it be there?

Also It feels like this list was sorted at some point. Maybe we should make it sorted again as it's in this more convenient form for that right now?

Would be easier to spot any missing things there if it's sorted.

u-boot/pkg.yaml
```diff
@@ -12,7 +12,7 @@ variant: scratch
 shell: /toolchain/bin/bash
 dependencies:
   - stage: base
-  - stage: libressl
+  - stage: openssl
```
frezbo, line 15: does this change affect this config: https://github.com/talos-systems/pkgs/blob/37e757d2919af029df784ff9ab5ad2fe8eb30e8a/u-boot/pkg.yaml#L67

Makefile
Unix4ever: > I don't see openssl in the list, should it be there?

Hmm, or it is part of `ipmitool`? :)

smira: yep, I think I missed `openssl`, and the list is (mostly) sorted, I will make sure it is sorted properly

Unix4ever (approved):

frezbo (approved):

PR #429, opened 2022-03-24
chore: org update

frezbo:
Org update

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 REGISTRY ?= ghcr.io
-USERNAME ?= talos-systems
+USERNAME ?= siderolabs
 SHA ?= $(shell git describe --match=none --always --abbrev=8 --dirty)
 TAG ?= $(shell git describe --tag --always --dirty)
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -96,4 +96,4 @@ kernel-%: ## Updates the kernel configs: e.g. make kernel-olddefconfig; make ker
 
 .PHONY: conformance
 conformance: ## Performs policy checks against the commit and source code.
-	docker run --rm -it -v $(PWD):/src -w /src ghcr.io/talos-systems/conform:v0.1.0-alpha.22 enforce
+	docker run --rm -it -v $(PWD):/src -w /src ghcr.io/siderolabs/conform:latest enforce
```
rgl, line 99: was changing to the `latest` version really intended?

frezbo: `conform` is for CI tests, shouldn't hurt :sweat_smile:

frezbo: on second thoughts since this is a backport makes sense to only change the org name, thanks for the pointer :+1:

smira (approved):

PR #445, opened 2022-04-11
chore: add tegra-drivers

frezbo:
Add new tegra OSS drivers


frezbo:
Still doesn't build, pushed to not lose track

frezbo:
These drivers don't support Tegra210 (Jetson Nano's) as of now, so closing.

PR #475, opened 2022-05-12
feat: add nvidia open gpu kernel modules

frezbo:
Add NVIDIA open gpu kernel modules

Ref: https://developer.nvidia.com/blog/nvidia-releases-open-source-gpu-kernel-modules/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,6 +50,7 @@ TARGETS = \
 	linux-firmware \
 	lvm2 \
 	musl \
+	nvidia-open-gpu-kernel-modules \
 	open-iscsi \
 	open-isns \
 	openssl \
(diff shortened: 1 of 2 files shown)


frezbo:
@smira i think we can create a system extension based on this released container image (we might need a kernel version constraint in the extension spec). I'm also not sure how to handle if the same kernel version is re-built as the signing key changes)

smira:
@frezbo I think we can do that easily if we pull exact `PKGS` version matching the release of Talos. So system extension will work exactly with the version of Talos it was built against. 

In other words, this package `nvidia-open-gpu-kernel-module` should match exactly `kernel` tag.

smira:
Or what this means we should rebuild/bump extensions for each Talos release.

frezbo:
> In other words, this package nvidia-open-gpu-kernel-module should match exactly kernel tag.

yeh, i was wondering the same as talos could have a pkgs bump if some other component got update

smira (approved):

PR #482, opened 2022-05-16
feat(kernel): build kernel with BTF enabled

charlie-haley:
This PR enables BTF support when building the kernel. Accompanying PR for adding build dependencies here https://github.com/siderolabs/tools/pull/197


charlie-haley:
Taking this out of draft, will retry the build once this is merged https://github.com/siderolabs/tools/pull/210

smira:
`tools` changes were merged, so to use new `tools` image please update `Pkgfile` with new reference to `tools` (new image tag is `v1.2.0-alpha.0-7-gac357ec`)

smira:
this doesn't seem to work as `make kernel-olddefconfig` removes the new option

smira:
e.g. our kernel enables https://cateee.net/lkddb/web-lkddb/GCC_PLUGIN_RANDSTRUCT.html and it's not compatible with https://cateee.net/lkddb/web-lkddb/DEBUG_INFO_BTF.html

plus it doesn't look pahole is detected: https://cateee.net/lkddb/web-lkddb/PAHOLE_VERSION.html

charlie-haley:
> e.g. our kernel enables https://cateee.net/lkddb/web-lkddb/GCC_PLUGIN_RANDSTRUCT.html and it's not compatible with https://cateee.net/lkddb/web-lkddb/DEBUG_INFO_BTF.html
> 
> plus it doesn't look pahole is detected: https://cateee.net/lkddb/web-lkddb/PAHOLE_VERSION.html

It looks like there's a few BTF conflicts, I'll play around locally and see if I can get a build working

rwunderer:
@charlie-haley I've found that at least for aquasec tracee to work, all of the following additional options have to be set:

```
CONFIG_TASKS_RUDE_RCU=y
CONFIG_KPROBES=y
CONFIG_OPTPROBES=y
CONFIG_KPROBES_ON_FTRACE=y
CONFIG_KRETPROBES=y
CONFIG_DEBUG_INFO=y
CONFIG_DEBUG_INFO_DWARF4=y
CONFIG_DEBUG_INFO_BTF=y
CONFIG_PAHOLE_HAS_SPLIT_BTF=y
CONFIG_DEBUG_INFO_BTF_MODULES=y
CONFIG_FUNCTION_TRACER=y
CONFIG_FUNCTION_GRAPH_TRACER=y
CONFIG_DYNAMIC_FTRACE=y
CONFIG_DYNAMIC_FTRACE_WITH_REGS=y
CONFIG_DYNAMIC_FTRACE_WITH_DIRECT_CALLS=y
CONFIG_DYNAMIC_FTRACE_WITH_ARGS=y
CONFIG_FTRACE_SYSCALLS=y
CONFIG_KPROBE_EVENTS=y
CONFIG_FTRACE_MCOUNT_RECORD=y
CONFIG_FTRACE_MCOUNT_USE_CC=y
CONFIG_FUNCTION_ERROR_INJECTION=y
```

Also `CONFIG_GCC_PLUGIN_RANDSTRUCT` needs to be unset as Andrey pointed out:

```patch
@@ -788,8 +792,7 @@ CONFIG_ARCH_HAS_GCOV_PROFILE_ALL=y
 CONFIG_HAVE_GCC_PLUGINS=y             
 CONFIG_GCC_PLUGINS=y                   
 CONFIG_GCC_PLUGIN_LATENT_ENTROPY=y                        
-CONFIG_GCC_PLUGIN_RANDSTRUCT=y       
-# CONFIG_GCC_PLUGIN_RANDSTRUCT_PERFORMANCE is not set               
+# CONFIG_GCC_PLUGIN_RANDSTRUCT is not set
 # end of General architecture-dependent options 
```

see also: https://github.com/siderolabs/pkgs/compare/master...caprisys:pkgs:feat/enable-btf

NB: `CONFIG_DEBUG_INFO_DWARF4=y` is especially important as the kernel will compile fine without it, but then containerd and cgroup2 will throw errors like 

```
in-kernel BTF is malformed
```

and in effect the whole system will be broken and unable to run any containers.

charlie-haley:
> @charlie-haley I've found that at least for aquasec tracee to work, all of the following additional options have to be set:
> 
> ```
> NB: `CONFIG_DEBUG_INFO_DWARF4=y` is especially important as the kernel will compile fine without it, but then containerd and cgroup2 will throw errors like
> 
> ```
> in-kernel BTF is malformed
> ```
> 
> and in effect the whole system will be broken and unable to run any containers.

Thanks for the pointers @rwunderer, looks we've got a successful build 🎉

kernel/build/config-amd64
```diff
@@ -786,10 +790,7 @@ CONFIG_ARCH_HAS_GCOV_PROFILE_ALL=y
 # end of GCOV-based kernel profiling
 
 CONFIG_HAVE_GCC_PLUGINS=y
-CONFIG_GCC_PLUGINS=y
```
smira, line 789: I think this is too much, we need to keep GCC_PLUGINS enabled, while (probably) disabling only the `CONFIG_GCC_PLUGIN_RANDSTRUCT=y`. Other gcc plugins provide meaningful security protection features which we shouldn't be dropping

charlie-haley: Sorry about the spam, accidentally pushed, you can cancel these 2 builds https://ci.dev.talos-systems.io/siderolabs/pkgs/1838
https://ci.dev.talos-systems.io/siderolabs/pkgs/1837

smira:
I updated the `config-amd64` to re-enable back some GCC_PLUGINS and removing some changes which seemed to be wrong

frezbo:
how much the vmlinuz size increases by this change?

frezbo (approved):

smira:
> how much the vmlinuz size increases by this change?

good question... I need to build it

smira (approved):
building the amd64 kernel to check the size

smira:
ok, the size diff is (expectedly) pretty big:

```
before: 15873024
after: 18801088
```

smira:
> it's +18%

that's for amd64, arm64 might be even bigger... as it's uncompressed

frezbo:
> > it's +18%
> 
> that's for amd64, arm64 might be even bigger... as it's uncompressed

wow, but if this eventually gives us ability to compile bpf programs on the fly, it's worth the tradeoff

PR #489, opened 2022-05-25
chore: stable source date epoch

frezbo:
Use the timestamp from repo initial commit as `SOURCE_DATE_EPOCH`

Bump to tools built with stable [`SOURCE_DATE_EPOCH`](https://github.com/siderolabs/tools/pull/198)


smira (approved):
can you please do the same for `toolchain`, `tools`, `extras`? (and probably `extensions`?)

PR #522, opened 2022-06-22
chore: enable `RANDOM_TRUST_BOOTLOADER` by default

frezbo:
Ref: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=846bb97e131d7938847963cca00657c995b1fce1


andrewrynhard:
It is possible we don't need to trust the CPU anymore. This was an older workaround for some changes in Linux that caused entropy generation to be super slow. It is possible that it fixed by now.

frezbo:
> 

yeh, the plan is to have it enabled in the kernel config and drop from the kernel cmdline in talos

smira (approved):

PR #549, opened 2022-08-12
revert: feat(kernel): build kernel with BTF enabled

smira:
This seems to cause kernel lockups after kexec.

This reverts commit 1ee594a807f09e5c48ed503feb9ae56caa19f391.


Unix4ever (approved):

smira:
@charlie-haley @rwunderer we're going to try reverting the BTF change unfortunately. it seems to be causing some lockups after kexec. If this change fixes that, we'll investigate what's up, but we need stable kernel before the next Talos release.

PR #561, opened 2022-08-22
fix: nvidia oss pkg name

frezbo:
Fix conflicting `pkg` and `extensions` image name.


frezbo:
Extensions image was also named the same https://github.com/siderolabs/extensions/blob/main/nvidia-gpu/nvidia-modules/pkg.yaml#L1

smira (approved):

PR #570, opened 2022-08-26
feat: enable renovate bot

frezbo:
Enable renovate dependency updates.


containerd/pkg.yaml
```diff
@@ -22,7 +22,7 @@ steps:
         export CGO_ENABLED=1
         export PATH=${PATH}:${TOOLCHAIN}/go/bin
         export BUILDTAGS='seccomp no_aufs no_btrfs no_devmapper no_zfs'
-        make bin/containerd bin/containerd-shim bin/containerd-shim-runc-v2 VERSION=v1.6.8 REVISION=9cd3357b7fd7218e4aec3eae239db1f68a5a6ec6
```
smira, line 25: I guess we need to handle `REVISION` here and for the `runc`

frezbo: that needs a manual update for now, either renovate doesn't work with git refs or I couldn't get it to work :sweat_smile:

smira: should we at least pull it out as a variable? I guess we'll forget about it, as everything else is a variable?

frezbo: yes, makes sense, also I might have found the correct example, I'll test and update

frezbo: fixed

smira (approved):

PR #578, opened 2022-09-05
chore: disable irc in conntrack

frezbo:
Talos users probably have no reason to use irc. Disable IRC in conntrack
to prevent issues like [NAT slipstreaming](https://dgl.cx/2022/08/nat-again-irc-cve-2022-2663)


smira (approved):

frezbo:
Also fixes an issue with xz download url

PR #580, opened 2022-09-11
feat: added gasket driver package

ammmze:
This driver enables the use of [Coral.ai](https://coral.ai/)'s PCIe TPU modules.

Signed-off-by: Branden Cash <203336+ammmze@users.noreply.github.com>

Required-by: siderolabs/extensions#69

### DIFF
--- a/gasket-driver/pkg.yaml
+++ b/gasket-driver/pkg.yaml
@@ -1,0 +1,37 @@
+name: gasket-driver-pkg
+variant: scratch
+shell: /toolchain/bin/bash
+dependencies:
+  - stage: kernel-build
+steps:
+  - sources:
+      - url: https://github.com/google/gasket-driver/archive/{{ .gasket_driver_ref }}.tar.gz
+        destination: gasket-driver.tar.gz
+        sha256: "{{ .gasket_driver_sha256 }}"
+        sha512: "{{ .gasket_driver_sha512 }}"
+    env:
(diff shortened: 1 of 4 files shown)


gasket-driver/pkg.yaml
```diff
@@ -0,0 +1,36 @@
...
+      - |
+        mkdir -p /rootfs/lib/modules/$(cat /src/include/config/kernel.release)/
+        
+        # create these so we don't get warnings when depmod runs
+        touch /rootfs/lib/modules/$(cat /src/include/config/kernel.release)/modules.order /rootfs/lib/modules/$(cat /src/include/config/kernel.release)/modules.builtin
+
+        make -C /src M=$(pwd)/src modules_install INSTALL_MOD_PATH=/rootfs
+
+        # remove these as we don't actually need/want them
+        rm /rootfs/lib/modules/$(cat /src/include/config/kernel.release)/modules.order /rootfs/lib/modules/$(cat /src/include/config/kernel.release)/modules.builtin
```
frezbo, line 33: sure this is not needed? the kmod library usually finds drivers from these files I believe

ammmze: From what I saw they remained empty files. Though I really only looked at modules.builtin. Initially when I had them still in there and installed the extension it overwrote the modules.builtin with an empty file. The modules seem to load just fine regardless of whether the file is empty or populated.

frezbo (approved):

smira (approved):

PR #585, opened 2022-09-13
chore: update deps

frezbo:
This PR addresses the following:

- Update deps.
- Add a `test` target to test reproducibility easily via `make rebuild-test`.
Currently `kernel` and `nvidia-open-gpu-kernel-modules` are not re-producible due to using a throw away kernel signing key.
- set `buildid` to empty for runc build for it to be reproducible, ref:
    - https://words.filippo.io/reproducing-go-binaries-byte-by-byte/
    - https://shibumi.dev/posts/day-in-the-life-of-a-package-maintainer-reproducible-go-packages/
- fix reproducibility of `flannel-cni`
- `ipmitool` might not be always re-producible due to the fact that it
  downloads a IANA database on build, so if that files changes
`diffoscope` would show it as a diff, the binaries are completely
reproducible.
- Add a check to verify out-of-tree kernel modules are signed
- Remove iscsi build from `pkgs`. iscsi-tools is now available as a
  system extension.


frezbo:
`pkgs` is also fully reproducible.

test/pkg.yaml
```diff
@@ -0,0 +1,43 @@
...
+  - stage: openssl
+  - stage: raspberrypi-firmware
+  - stage: runc
+  - stage: socat
+  - stage: syslinux
+  - stage: u-boot
+  - stage: util-linux
+  - stage: xfsprogs
+finalize:
+  - from: /toolchain
```
smira, line 42: this should make it empty? as should be nothing in `/toolchain`

test/pkg.yaml
```diff
@@ -0,0 +1,43 @@
+name: test
+variant: scratch
+shell: /toolchain/bin/bash
+dependencies:
+  - stage: base
```
smira, line 5: `base` is never published iirc

test/pkg.yaml
frezbo: ahhh :facepalm:

smira (approved):

test/pkg.yaml
frezbo: i wonder why diffoscope didn't complain

frezbo: > i wonder why diffoscope didn't complain

nvm, both targets were empty

smira (approved):

PR #593, opened 2022-09-23
Add a u-boot build of "rpi_arm64_defconfig"

myndzi:
This configuration uses the device-tree provided by the firmware rather than embedding one; it's the only configuration that lists Compute Module 4 support. See https://github.com/ARM-software/arm-trusted-firmware/blob/master/docs/plat/rpi4.rst


myndzi:
Not sure if I actually have to set up a GPG identity to contribute, or if those build checks are for contributor signoff...

frezbo:
> but I don't know what is required to pass the gpg-identity check

that check will pass once we force push, we require the commits to be signed by one of the devs

myndzi:
> > but I don't know what is required to pass the gpg-identity check
> 
> that check will pass once we force push, we require the commits to be signed by one of the devs

Ah, I kind of thought that would be the case, but I could also see my way to reasoning for "contributors need to do these things" too, so I spent some time figuring out how to get set up 🤷

Perhaps consider updating the PR template to make a note of this (which checks should pass, and which the contributor doesn't need to worry about)?

frezbo (approved):

PR #598, opened 2022-09-28
chore: bump kernel to 5.15.71

frezbo:
Bump kernel to 5.15.71


.dockerignore
```diff
@@ -0,0 +1,3 @@
+.git
```
smira, line 1: I'm not sure if this useful... as it's `bldr` based, I believe this has no effect. (and `bldr` never loads whole directory)

smira (approved):

PR #606, opened 2022-10-12
feat: re-enable build kernel with BTF enabled

smira:
This reverts commit 165c2786f844647ecd11c39dcf6db4cfed3ad096.


frezbo (approved):

smira:
I verified manually that kexec works from pre-BTF kernel to post-BTF kernel now, so we can give this another spin.

PR #611, opened 2022-10-20
chore: update dependencies

TimJones:
Updates:
* Linux kernel 5.15.74
* Linux firmware 20221012

Reverts
* openssl 1.1.1q (see https://mta.openssl.org/pipermail/openssl-announce/2022-October/000237.html)

Signed-off-by: Tim Jones <tim.jones@siderolabs.com>

### DIFF
--- a/kernel/build/config-arm64
+++ b/kernel/build/config-arm64
@@ -1,6 +1,6 @@
 #
 # Automatically generated file; DO NOT EDIT.
-# Linux/arm64 5.15.73 Kernel Configuration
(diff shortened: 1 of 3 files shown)


kernel/build/config-arm64
```diff
@@ -1,6 +1,6 @@
 #
```
frezbo, line 1: this is `make olddefconfig` right? just to be safe

TimJones: `make kernel-olddefconfig`

smira (approved):

PR #614, opened 2022-10-25
chore: update packages version

Unix4ever:
Via renovate.

Signed-off-by: Artem Chernyshev <artem.chernyshev@talos-systems.com>

### DIFF
--- a/Pkgfile
+++ b/Pkgfile
@@ -11,10 +11,10 @@ vars:
   cni_sha512: 03da31caee5f9595abf65d4a551984b995bc18c5e97409549f08997c5a6a2b41a8950144f8a5b4f810cb401ddbe312232d2be76ec977acf8108eb490786b1817
 
   # renovate: datasource=github-tags depName=containerd/containerd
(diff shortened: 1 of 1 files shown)


Pkgfile
```diff
@@ -164,9 +164,9 @@ vars:
   arm_trusted_firmware_sha512: 3e1e09bcc69a0f080f91b94d9c8da6dd0b27b86292599c280438d7b6f594017a896af7cfc58d9d22053f77b5daa2faecd51a59505c755eac8493e4e0fd1e9203
 
   # renovate: datasource=github-tags extractVersion=^v(?<version>.*)$ depName=u-boot/u-boot
-  uboot_version: 2022.07
```
smira, line 167: please don't bump u-boot, I think Noel said we need to update patches which we don't have yet

Unix4ever: ah, ok

Unix4ever: reverted that

smira (approved):

PR #624, opened 2022-12-05
Add nitro enclave support in kernel

lrvick:
Want to start be able to schedule nitro enclaves on Talos OS, and this seems to be the first blocker.

End goal is to essentially replicate this setup in Talos: https://superorbital.io/journal/aws-nitro-enclaves-in-k8s-pods/


smira (approved):

frezbo:
todo: check if similar arm64 config is supported

frezbo:
> todo: check if similar arm64 config is supported

This in only an amd64 setting, approving

PR #633, opened 2022-12-19
fix: patch ipmitool IANA URL

smira:
See https://github.com/siderolabs/sidero/issues/1013 
See https://github.com/ipmitool/ipmitool/issues/377

### DIFF
--- a/ipmitool/patches/iana-url.patch
+++ b/ipmitool/patches/iana-url.patch
@@ -1,0 +1,15 @@
+https://github.com/ipmitool/ipmitool/commit/1edb0e27e44196d1ebe449aba0b9be22d376bcb6.diff
+
+diff --git a/Makefile.am b/Makefile.am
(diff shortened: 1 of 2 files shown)


frezbo (approved):
Nice, I though it was the build process that was broken :sweat_smile:

smira:
> Nice, I though it was the build process that was broken sweat_smile

:shrug: the released version is broken, but there's no new release :facepalm:

frezbo:
> > Nice, I though it was the build process that was broken sweat_smile
> 
> shrug the released version is broken, but there's no new release facepalm

should we add a `test` section in `pkg.yaml` that just run `ipmitool` to make sure it works? :thinking:

smira:
I think a test requires at least some valid IPMI target. But in general I agree - we should test more in `pkgs` if possible

Unix4ever (approved):

PR #636, opened 2022-12-20
fix: use proper containerd version tag

smira:
Tag got out of sync.


frezbo (approved):
glad we have tests :smile:

PR #649, opened 2023-01-11
backport: feat: Linux 5.15.86, containerd 1.6.15

smira:
See https://github.com/containerd/containerd/releases/tag/v1.6.15


frezbo (approved):
should we do the new go version?

smira:
> should we do the new go version?

looks like no security fixes, so I would rather not do it, as we don't see any immediate problem fixed with Go 1.19.5

PR #665, opened 2023-01-30
chore: bump deps and disable un-needed kconfig

frezbo:
Bump deps

Disable support for HAM radio and Wireless.


smira (approved):

Unix4ever (approved):

james-callahan:
Why disable wireless?
Seems like it would be useful for e.g. raspberry pi users.

Could it be made an optional module instead?

frezbo:
> Why disable wireless?
> Seems like it would be useful for e.g. raspberry pi users.

Wireless support was never enabled in talos, this was just disabling some part of it which was missed. If anyone needs wireless drivers it should be a system extension

PR #675, opened 2023-02-15
chore: bump deps

frezbo:
Bump kernel to 6.1.12
Bump Go to 1.20.1
Bump cacert to 20230110

Also fixes base to copy libstdc++ libs.

### DIFF
--- a/base/pkg.yaml
+++ b/base/pkg.yaml
@@ -9,6 +9,7 @@ steps:
       - |
         cp -R /toolchain/lib/gcc /lib
         cp -R /toolchain/lib/libgcc* /lib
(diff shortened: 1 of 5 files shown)


base/pkg.yaml
```diff
@@ -9,6 +9,7 @@ steps:
       - |
         cp -R /toolchain/lib/gcc /lib
         cp -R /toolchain/lib/libgcc* /lib
+        cp -R /toolchain/lib/libstdc* /lib
```
frezbo, line 12: this took me a while to figure out :sweat_smile:

smira (approved):

PR #684, opened 2023-03-10
feat: update to Go 1.19.7

smira:
Also update Linux to 5.15.98.


frezbo:
> Also update Linux to 5.15.98.

seems 5.15.99 is right away :joy:

PR #688, opened 2023-03-19
Enabled Hyper-V dynamic memory driver

SixFive7:
Enabling this driver has no effect for default Hyper-V users. But it does allow enabling dynamic memory for users that opt in to that feature.


frezbo:
could you do a git signoff `git commit -s --amend --no-edit` and `git push --force-with-lease` should do

smira:
do we need to do same for `config-arm64` ?

SixFive7:
> could you do a git signoff `git commit -s --amend --no-edit` and `git push --force-with-lease` should do

Not sure if this is something that you want me to do, but I can't do that with this PR as it was made with the Github web interface. It was just a one-liner, sorry. I can only write in this comment that I will happily sign of on it ;-)

SixFive7:
> do we need to do same for `config-arm64` ?

A quick search delivers me the news that Hyper-v is actually working on arm hardware nowadays. Did not know this. As a result it should indeed also be enabled for `config-arm64`.

smira (approved):

PR #691, opened 2023-03-28
Default to enable unsupported GPUs

markbaumgarten:
Wondering if this pr might be useful. Will do some testing tomorrow, hoping this will fix #7008


markbaumgarten:
I wonder if changing this default flag will affect anyone other than those who, like me, don't have a "datacenter GPU".

The hoops one have to jump through to get socalled "unsupported" GPUs working are a real pita.

However I am really pleased with Talos and really appreciate all of your efforts in making such a nice K8sOS, so I will jump through hoops if I have to :-) . Thanks!

frezbo:
> I wonder if changing this default flag will affect anyone other than those who, like me, don't have a "datacenter GPU".

I'm not completely sure of it, so better to fix talos so that module parameters are applied correctly even if module is already loaded.

PR #694, opened 2023-03-31
feat: use cache kernel build output

frezbo:
Use cache for kernel build output and re-use in downstream steps.

### DIFF
--- a/kernel/build/pkg.yaml
+++ b/kernel/build/pkg.yaml
@@ -7,15 +7,17 @@ steps:
   - env:
       CARCH: {{ if eq .ARCH "aarch64"}}arm64{{ else if eq .ARCH "x86_64" }}amd64{{ else }}unsupported{{ end }}
       ARCH: {{ if eq .ARCH "aarch64"}}arm64{{ else if eq .ARCH "x86_64" }}x86_64{{ else }}unsupported{{ end }}
+      KBUILD_OUTPUT: /src/build
+    cache: /src/build
     prepare:
       - |
         cd /src
-        cp -v /pkg/config-${CARCH} .config
+        cp -v /pkg/config-${CARCH} ${KBUILD_OUTPUT}/.config
         cp -v /pkg/certs/* certs/
     build:
       - |
         cd /src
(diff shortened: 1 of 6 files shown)


kernel/build/pkg.yaml
```diff
@@ -7,15 +7,17 @@ steps:
...
       - |
         cd /src
-        cp -v /pkg/config-${CARCH} .config
+        cp -v /pkg/config-${CARCH} ${KBUILD_OUTPUT}/.config
         cp -v /pkg/certs/* certs/
     build:
       - |
         cd /src
-        python3 /toolchain/kconfig-hardened-check/bin/kconfig-hardened-check -c .config -m json | python3 /pkg/scripts/filter-hardened-check.py
+        python3 /toolchain/kconfig-hardened-check/bin/kconfig-hardened-check -c ${KBUILD_OUTPUT}/.config -m json | python3 /pkg/scripts/filter-hardened-check.py
```
smira, line 20: my concern is how can we be sure that the cache gets properly invalidated? I guess in default `make`, it's all timestamp-based, so if we have two kernel builds from different branches, it may very well poison the cache?

frezbo: The `src` folder gets changed, but the kernel build objects can still be re-used. I don;t want this going in release-1.4 anyways, targetting for 1.5

frezbo: I think the kernel build is intelligent enough to detect that, based on the source. At least when building locally, but I'ven't tried changing versions :sweat_smile: . I'll test locally later

smira: if there's some smartness, let's find some proof on that :) it's definitely not in `make` itself

pl4nty:
I've been using this for a few months on RISC-V, haven't tested invalidation across kernel versions yet but I've run into corruption issues. A misconfigured dependent pkg can write to the cache and corrupt it for other pkgs. This doesn't invalidate the cache, and I've had to delete it manually.

Would a `cachePaths[].readOnly` property make sense? Happy to send a bldr PR if so

PR #705, opened 2023-04-10
chore: copy over the kernel signing public key

frezbo:
Copy over the kernel signing public key so that it can be used to verify modules are signed by the correct key. The signing key is copied over as otherwise in-order to extract the signing key we either need the uncompressed `vmlinux` (which is super huge to copy over) or need the `System.map` file along with `vmlinuz` and a perl script to extract it. It's just easier to just copy over the `signing_key.x509` file.

### DIFF
--- a/kernel/kernel/pkg.yaml
+++ b/kernel/kernel/pkg.yaml
@@ -12,6 +12,8 @@ steps:
 
         mkdir -p /rootfs/boot
         mkdir -p /rootfs/dtb
+        mkdir -p /rootfs/certs
+        cp /src/certs/signing_key.x509 /rootfs/certs/signing_key.x509
         case $ARCH in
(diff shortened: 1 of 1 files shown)


DmitriyMV (approved):

smira (commented):
is it really _public_ key? don't we need to extract the public part from it?

frezbo:
> is it really _public_ key? don't we need to extract the public part from it?

yes, https://github.com/siderolabs/pkgs/blob/main/kernel/build/config-amd64#L5450 is the private key. `signing_key.x509` is the public part in DER format

frezbo:
https://www.kernel.org/doc/html/v4.15/admin-guide/module-signing.html#configuring-module-signing

PR #713, opened 2023-04-26
feat: nfsd support as module package

TMaYaD:
Build nfsd support as module package. This allows to build an nfsd extension to be used with projects like [openebs/dynamic-nfs-provisioner](https://github.com/openebs/dynamic-nfs-provisioner).

Talos can load it via machine-config:
```yaml
machine:
    # Configures the kernel.
    kernel:
        # Kernel modules to load.
        modules:
            - name: nfsd # Module name.
```


nfsd/nfsd.config
```diff
@@ -0,0 +1,9 @@
+CONFIG_NFSD=m
```
smira, line 1: should we simply build `nfsd` as a module as part of the main kernel build @frezbo ?

frezbo: I think we update the the Kconfig and only build the modules that we ship by default in talos, then these extra can be build as separate modules shipped as extensions, so we don't increase initramfs size

smira: ah, okay... makes sense.

TMaYaD: That's the same approach as I'm taking here. Right?

LMK if there is anything else needed to get this approved and merged in.

frezbo: > That's the same approach as I'm taking here. Right?

meant adding to the actual kernel config under `kernel/build/config-<arch>`, I'll try to experiment and see if that approach works for our use case. Might take a while

frezbo: probably the latter, I'll have to experiment to see what works

TMaYaD: I couldn't currently figure out where `kernel` is being used to build `installer`. If you can point me, I'll be more than happy to run a few experiments of my own.

frezbo: https://github.com/siderolabs/talos/blob/main/Dockerfile#L77-L99 you can update this with your kernel image and do `make installer PUSH=true`

frezbo: quick question: does this work without rpc service running?

frezbo: @TMaYaD does this require rpc service? Else I can fix this up and get this merged.

TMaYaD: I worked on this a while back and don't remember how I got this working. I'll check and get back to you.

tyriis:
any chance to get this pushed?

AngellusMortis:
With 1.8 having a lot of focus on improving volume management, this might be a great thing to try to get in.

cm-enfuse:
I am attempting to update this PR. I have managed to build nfsd on the following branch:
https://github.com/cm-enfuse/talos-pkgs/tree/nfsd

I have managed to get an image built that has the module available, but, since I didn't sign it(?) it refuses to load. Would it be possible to get a pointer to how to do that so I can verify the module works as expected?

`host: kern:  notice: [2025-02-28T23:03:43.913581228Z]: Loading of module with unavailable key is rejected
`
`
host: user: warning: [2025-02-28T23:03:43.914382228Z]: [talos] controller failed {"component": "controller-runtime", "controller": "runtime.KernelModuleSpecController", "error": "error loading module \"nfsd\": load nfsd failed: key was rejected by service"}
`

In regard to rpc, only nfs v4 can run without rpc. If you take a look at the [nfsd.conf](https://github.com/cm-enfuse/talos-pkgs/blob/nfsd/nfsd/nfsd.config) file, it should be the only version that is being built and should only enable v4. If I can get the module to load, I will confirm as it is my assumption that there is no currently running rpc service per the pervious comments.

Thank you for any help! If this was the wrong place to put this, please let me know and I will delete and move to the right location. 

The other relevant branches for the rest of the stack:
https://github.com/cm-enfuse/talos-extensions/tree/nfsd
https://github.com/cm-enfuse/talos/tree/nfsd

smira:
@cm-enfuse thanks for taking this effort further, we have a guide on building Talos with a custom kernel: https://www.talos.dev/v1.9/advanced/customizing-the-kernel/

You need to build both kernel & modules to get signatures in sync.

You can add nfsd module(s) into `hack/modules-amd64.txt` in Talos to get it shipped as part of the initial boot assets, so it's easy to test it.

cm-enfuse:
@smira Many thanks, I hadn't built the imager and that was the crux of the problem.

With the above changes in place, I was able to successfully run an nfs server container, with privileged, and externally mount it and verify a file was written.

How would you like me to proceed  for submitting for review?

cm-enfuse:
Created new PR. Thank you for your help; let me know what I need to do in order to get past all the failing checks. I apologize for thrash on that part.
https://github.com/siderolabs/pkgs/pull/1174

smira:
@cm-enfuse thank you, I will close this PR, and make sure your PRs are merged for Talos 1.10!

PR #719, opened 2023-04-28
chore: selectively build modules

frezbo:
This change allows the Talos kernel to be shipped with modules that are currently being built as modules (mellanox and virtio). Any extra module can still be enabled in kernel config but won't be shipped by default. They can be built as extensions and enabled separately.

### DIFF
--- a/kernel/build/pkg.yaml
+++ b/kernel/build/pkg.yaml
@@ -20,7 +20,11 @@ steps:
         cd /src
 
         make -j $(nproc)
-        make -j $(nproc) modules
+        {{ if eq .ARCH "aarch64"}}
+        make -j $(nproc) M=drivers/net/ethernet/mellanox
+        {{ else }}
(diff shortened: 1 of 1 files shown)


smira (commented):
should we go one step further, and build all modules here, and simply copy only the ones we need to ship with Talos by default?

frezbo:
> should we go one step further, and build all modules here, and simply copy only the ones we need to ship with Talos by default?

but then the depmod file would have references to modules not shipped by default, unless we run `depmod` before the final stage

smira:
> > should we go one step further, and build all modules here, and simply copy only the ones we need to ship with Talos by default?
> 
> but then the depmod file would have references to modules not shipped by default, unless we run `depmod` before the final stage

yep, exactly, just run `depmod` as part of `initramfs` build in Talos

frezbo:
This is not needed 

> should we go one step further, and build all modules here, and simply copy only the ones we need to ship with Talos by default?

This PR can be closed as changes are going to be in Talos as per the discussion above.

PR #722, opened 2023-05-08
feat: update Linux 6.1.27, containerd 1.6.21

smira:
Bumping to the latest.


frezbo (approved):
shouldn;t we pull in go too?

smira:
> shouldn;t we pull in go too?

I think it's already there

frezbo:
i think i only did a PR to tools

PR #740, opened 2023-06-08
fix: copy missing `modules.*` files

frezbo:
Copy missing `modules.order`, `modules.builtin` and `modules.builtin.modinfo` files so tools can read them.

### DIFF
--- a/usb-modem-drivers/files/modules.txt
+++ b/usb-modem-drivers/files/modules.txt
@@ -1,3 +1,6 @@
+modules.order
+modules.builtin
+modules.builtin.modinfo
 kernel/drivers/net/usb/dm9601.ko
 kernel/drivers/net/usb/rndis_host.ko
 kernel/drivers/net/usb/smsc75xx.ko


usb-modem-drivers/pkg.yaml
```diff
@@ -12,6 +12,11 @@ steps:
       - |
         export KERNELRELEASE="{{ .linux_version }}-talos"
 
+        # we need these so tools relying on modules.builtin can find the modules statically compiled into kernel
```
smira, line 15: would it conflict with same files from `kernel`? would it be same?

frezbo: so depmod won;t generate these files, it has to come from base kernel, since `modules.builtin` contain the list of modules that is statically builtin

Also some tools also check this file to see if module is already loaded

frezbo: moved to the file list, since `install` command handles it better

smira (approved):

PR #746, opened 2023-06-22
feat: add alx drivers

rsmitty:
This PR adds support for more qualcomm atheros nics. See https://cateee.net/lkddb/web-lkddb/ALX.html for details on the actual drivers.


frezbo (approved):
is this needed at early boot? or can we do as module?

frezbo:
> is this needed at early boot? or can we do as module?

I meant udevd could (if the driver provides  pci id info) load this module automatically if hardware is present

kernel/build/config-amd64
```diff
@@ -2392,7 +2392,7 @@ CONFIG_NET_VENDOR_ATHEROS=y
 # CONFIG_ATL1 is not set
 # CONFIG_ATL1E is not set
 # CONFIG_ATL1C is not set
-# CONFIG_ALX is not set
```
smira, line 2395: if we build as modules, we should include them as the list of files we copy by default to `initramfs`

PR #749, opened 2023-07-02
feat: enable sriov

sergelogvinov:
* support HyperV sr-iov network
* suppprt SRIOV CNI plugin for amd64, arm64 already has it.

https://github.com/siderolabs/talos/issues/7415


frezbo (commented):
would building these as modules work?

smira:
> would building these as modules work?

I think base PCI_HYPERV can't be loaded dynamically by udev, and parts of bnx2 we already have built-in, so enabling one more piece doesn't sound like it should be a module?

smira (approved):

PR #752, opened 2023-07-08
Enable USB attached SCSI driver on x86 systems

### LINKED ISSUE #751: Enable USB Attached SCSI (UAS) driver
The UAS driver allows for much better read and write performance and overcomes shortcomings in the bulk only transfer mode around command queueing and streaming. 

I believe it is as simple as enabling https://github.com/siderolabs/pkgs/blob/43451e68a0ddf634b90c7c12cca9437faa52d183/kernel/build/config-amd64#L4152C9-L4152C9 and I will be trying to build a custom Talos image with this enabled.

You can confirm the driver is loaded with `dmesg | grep usbcore` where you see a line like

> usbcore: registered new interface driver uas

cehoffman:
This should close #751 

After a bit of issue following the directions to building a custom kernel I was able to verify UAS driver is working now.


smira (approved):
Thanks for your PR!

PR #762, opened 2023-07-22
feat: kvm as module package

kvaps:


kvaps:
Ah just noticed that `kvm` module already in-tree

```
# talosctl cat /lib/modules/6.1.28-talos/modules.builtin | grep kvm
kernel/arch/x86/kvm/kvm.ko
kernel/arch/x86/kvm/kvm-intel.ko
kernel/arch/x86/kvm/kvm-amd.ko
kernel/drivers/ptp/ptp_kvm.ko
```

PR #774, opened 2023-08-09
chore: enable pushing of non-free packages

smira:
At the moment it is kmod-nvidia.


.drone.yml
```diff
@@ -121,6 +121,7 @@ steps:
     commands:
       - docker login ghcr.io --username "$${GHCR_USERNAME}" --password "$${GHCR_PASSWORD}"
       - make PUSH=true
+      - make nonfree PUSH=true
```
frezbo, line 124: might be better to move to default list?

smira: I would keep them separate if we can... I think that makes it easier (mentally) to keep the distinction between the two

PR #785, opened 2023-08-15
add btrfs support

Gottox:
Part 1 of 2 to add a btrfs extension to talos. Part 2 will happen in the extension repository.

tracking issue: https://github.com/siderolabs/extensions/issues/199


kernel/build/config-amd64
```diff
@@ -5421,11 +5427,11 @@ CONFIG_CRYPTO_CRC32=y
 # Compression
 #
 CONFIG_CRYPTO_DEFLATE=y
-# CONFIG_CRYPTO_LZO is not set
+CONFIG_CRYPTO_LZO=y
```
frezbo, line 5430: i guess these cannot be built as modules?

Gottox: I chose to make it a builtin, as it is also a builtin in the arm64 kernel.

Gottox: Also I wasn't sure if I need to put them into its own dedicated package or if it's fine to ship it with the btrfs one.

Gottox: One issue comes up if build `CONFIG_LZO_DECOMPRESS=m` on amd64:

This option is already a builtin on arm64, therefore the list of modules to package differs on arm64 vs amd64.

* option 1: make it a builtin on both platforms
* option 2: is there a way to have platform specific module lists?

frezbo: the diff is probably an oversight, we can make it similar, if it's a dependency for btrfs and can be built as a module, let's do that

Gottox: something is using LZO in the arm-kernel... I don't have proper tooling to investigate, but if I apply the values from amd64 to arm64 and run `make oldconfig`, they get reenabled to `=y`.

Gottox: applied.

kernel/build/config-amd64
```diff
@@ -4941,7 +4941,13 @@ CONFIG_XFS_POSIX_ACL=y
 # CONFIG_XFS_WARN is not set
 # CONFIG_XFS_DEBUG is not set
 # CONFIG_GFS2_FS is not set
-# CONFIG_BTRFS_FS is not set
+CONFIG_BTRFS_FS=m
+CONFIG_BTRFS_FS_POSIX_ACL=y
```
frezbo, line 4945: is this required?

Gottox: The option does not affect the resulting kernel, the ACL support will be build into the module.

frezbo (approved):
Thank you :heart:

Gottox:
> Thank you ❤️

Thanks for the great review! I loved being challenged by your aspiration to minimalism!

Gottox:
Should I open another issue about my findings with amd vs arm kernels and their `CONFIG_LZO_DECOMPRESS` deviation?

frezbo:
> Should I open another issue about my findings with amd vs arm kernels and their `CONFIG_LZO_DECOMPRESS` deviation?

I kind of looked into it and it's seems to be the (I believe so, not completely sure) that arm kernels are not compressed. At least for the dependency that enables LZO_DECOMPRESS, but sometimes that's not the full story with Kconfig :sweat_smile: . It was interesting to see that just enabling btrfs doesn't need CONFIG_LZO to be enabled, but only to make both the configs to be in parity :shrug: . Sure please add another issue

PR #791, opened 2023-08-17
feat: add device-mapper package

ndbrew:
Add kernel modules for device-mapper.


ndbrew:
I decided to depend on the kernel-build using it as a cache to build the required modules in this package rather than modify the underlying kernel-build configs. I was unable to make this work due to make invalidating the kernel-build cache every time I built the package resulting in errors while loading the module. I assume that's what #694 aims to address?

To get it working locally I ended up making kernel-build a package in the makefile and pushing it to docker, then depending on that image instead of the stage. Works fine as long as this package and the kernel package use the same kernel-build cache.

frezbo:
> I decided to depend on the kernel-build using it as a cache to build the required modules in this package rather than modify the underlying kernel-build configs. I was unable to make this work due to make invalidating the kernel-build cache every time I built the package resulting in errors while loading the module. I assume that's what #694 aims to address?
> 
> To get it working locally I ended up making kernel-build a package in the makefile and pushing it to docker, then depending on that image instead of the stage. Works fine as long as this package and the kernel package use the same kernel-build cache.

buildkit needs some config to tell it to no do GC, https://github.com/siderolabs/build-container/blob/main/hack/scripts/buildkit.toml, I'll clean up the PR and include the config into the standard place.

Does this require any userspace binaries/daemon?

ndbrew:
> Does this require any userspace binaries/daemon?

Not for the piraeus-operator / linstor use case.

ndbrew:
I've updated the kernel configs and ran `kernel-olddefconfig`. These are the resulting changes.

frezbo:
I'll get back to this next week, apologies for the delay

frezbo (approved):

PR #793, opened 2023-08-18
feat: add applesmc package

buroa:
This adds the `applesmc` pkg to use in extensions.

Because of the nature of applesmc on T2 mac's, they are using patches currently.

Currently this targets the 6.1 (kernel) branch of https://github.com/t2linux/linux-t2-patches. I put the renovate comment in and a note to not bump it unless we start building with 6.2+, which then we would switch the reference.

I'm willing to keep the `applesmc` extension up to date as I depend on it.

Extension: https://github.com/siderolabs/extensions/pull/207

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,6 +72,7 @@ TARGETS = \
 #   kernel first, then packages in alphabetical order
 TARGETS += \
 	kernel \
+	applesmc-pkg \
 	btrfs-pkg \
 	drbd-pkg \
 	gasket-driver-pkg \
--- a/Pkgfile
+++ b/Pkgfile
@@ -192,5 +192,11 @@ vars:
   zfs_version: 2.1.12
   zfs_sha256: 64daa26aed3e12c931f6f4413d7527c4ebdb8da35416b356152b5f9fdd4c6e6d
   zfs_sha512: f48493a21883e441cda705fb085353bed033f1620a1d0f93069c345c76cf2c0759a2e6f7a80c47c9398e9878abfe1d90d931fe5ceaf2588770a71491a434631e
+
+  # DO NOT bump until we are on kernel version 6.2 or newer.
+  # renovate: datasource=git-refs versioning=git depName=https://github.com/t2linux/linux-t2-patches.git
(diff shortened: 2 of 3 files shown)


smira:
I'm not sure we're willing to accept any non-upstream kernel modules due to maintenance costs associated with that.

buroa:
> I'm not sure we're willing to accept any non-upstream kernel modules due to maintenance costs associated with that.

The module itself is upstreamed, however, it does not work with Apple T2 devices. I'm not sure why the patches have never been upstreamed but I know they have been submitted.

Maintaining this patch is quiet easy, for both of us, but like I said above -- I'm willing to help out here because I depend on this module in my Talos builds (and maybe others in the future).

Please let me know how to proceed here.

smira:
First of all, thanks for your effort in contributing this.

The problem I might see in the future is the following:

* as soon as we publish this module, some users might start using it and relying on it
* a new kernel drops in 6.1 (or the next LTS), and the patches don't apply in a clean way
* the kernel has to be updated (e.g. fixing some CVE), but the patches block it
* so we either have to hold on updating the kernel while putting some users at risk, or we would have to stop building this package, and some users might rely on it

It's always a balance between the two, and probably Apple T2 devices are not the most common hardware platform for Talos.

For exactly same reason we don't support SBCs which don't have upstream support in the kernel/u-boot, and we already have some u-boot patches which are extremely hard to manage.

We will discuss this issue internally as well.

smira:
@buroa getting back to this issue, we would rather not support this upstream Talos.

we are also looking into ways to make it easier to support out-of-tree kernel modules, potentially signed by other keys

buroa:
> @buroa getting back to this issue, we would rather not support this upstream Talos.
> 
> we are also looking into ways to make it easier to support out-of-tree kernel modules, potentially signed by other keys

If you can get it so we can ship modules (as extensions) that aren't signed by the main installer image's throwaway key, that would work perfectly for me. Maybe during a upgrade, Talos resigns them or something.

I'd definitely maintain this extension myself then :)

PR #798, opened 2023-09-04
chore: bump deps

frezbo:
Updates

| Package | Update | Change |
|---|---|---|
| git://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git | patch | `6.1.46` -> `6.1.51` |
| git://git.kernel.org/pub/scm/utils/util-linux/util-linux.git | patch | `2.39.1` -> `2.39.2` |
| https://github.com/google/gasket-driver.git | digest | `97aeba5` -> `8aca235` |
| https://github.com/ipxe/ipxe.git | digest | `c1834f3` -> `9e99a55` |

This also fixes the zfs dubious version being picked up as an update. (ZFS seems to tag `.99` patch version before an actual release).

Also moves `make kernel-olddefconfig` into `buildx` target to avoid loading into local docker.

### DIFF
--- a/kernel/build/config-amd64
+++ b/kernel/build/config-amd64
@@ -1,6 +1,6 @@
 #
 # Automatically generated file; DO NOT EDIT.
-# Linux/x86 6.1.46 Kernel Configuration
+# Linux/x86_64 6.1.51 Kernel Configuration
 #
 CONFIG_CC_VERSION_TEXT="gcc (GCC) 12.3.0"
 CONFIG_CC_IS_GCC=y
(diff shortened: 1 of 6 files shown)


kernel/build/config-amd64
```diff
@@ -1,6 +1,6 @@
 #
 # Automatically generated file; DO NOT EDIT.
-# Linux/x86 6.1.46 Kernel Configuration
+# Linux/x86_64 6.1.51 Kernel Configuration
```
frezbo, line 3: this changed from `x86` to `x86_64` since `CARCH` is set properly in the `kernl/build/pkg.yaml`

PR #804, opened 2023-09-13
feat: add binfmt_misc support

### LINKED ISSUE #803: binfmt_misc as module
Hello

Can we add the binfmt_misc as a module in the kernel
https://github.com/siderolabs/pkgs/blob/7d9e60e33fb602c81c61112f3557808a7064bb9a/kernel/build/config-amd64#L942

it helps to use Talos as build machine. https://github.com/multiarch/qemu-user-static

Thanks.

sergelogvinov:
This changes the kernel config to enable `binfmt_misc` support and adds a new package for the binfmt_misc kernel module.

I did not check all pipeline to add extensions to the Talos, but i hope i am on my way.

closes #803


binfmt-misc/pkg.yaml
```diff
@@ -0,0 +1,23 @@
+name: binfmt-misc-pkg
```
smira, line 1: we could probably skip this, the module will be part of the `kernel` pkg this way, but not added by default to Talos initramfs (that's where the extension would help)

smira: @frezbo please correct me if I'm wrong - do we actually need to a separate package here in `pkgs`, or we could simply strip out a single file from `kernel` for the extension?

frezbo: hmm, good point, we can definitelty do that from extensions and drop shipping `-pkg` suffixes images for in-tree modules

frezbo: created #805 to track

smira: ok, sounds good, I'll update this PR, thanks

smira (approved):

PR #808, opened 2023-09-15
feat: add Chelsio support

### LINKED ISSUE #807: Enable recent Chelsio NICs
These are fairly popular NICs for hyperconverged storage.

dhess:
Closes #807


dhess:
(I'm flying blind here as I'm not sure how the whole Talos kernel/image/installer building process works. Hopefully this compiles straightforwardly.)

frezbo:
does this need to be part of standard talos or can be brought as an extension?

dhess:
In my case, the bare metal host only has a Chelsio NIC connected to the switch, so I can't even get it on the network to bring it out of maintenance mode & bootstrap. Wouldn't that rule out an extension?

frezbo:
> In my case, the bare metal host only has a Chelsio NIC connected to the switch, so I can't even get it on the network to bring it out of maintenance mode & bootstrap. Wouldn't that rule out an extension?

yes, but you could still build a custom iso with the extensions, anyways i guess this can go in Talos for now

frezbo:
Is this needed? `CHELSIO_INLINE_CRYPTO`?

dhess:
> Is this needed? `CHELSIO_INLINE_CRYPTO`?

Yes, it appears so, and `CONFIG_TLS` as well, which I assumed would already be set in the Talos kernel config, but appears not to be. I guess I'll have to fire up `menuconfig` for the first time in 10 years and do this properly. :) Give me a few minutes and I'll update the PR.

frezbo:
> > Is this needed? `CHELSIO_INLINE_CRYPTO`?
> 
> Yes, it appears so, and `CONFIG_TLS` as well, which I assumed would already be set in the Talos kernel config, but appears not to be. I guess I'll have to fire up `menuconfig` for the first time in 10 years and do this properly. :) Give me a few minutes and I'll update the PR.

no worries,I already started on it

frezbo:
> > Is this needed? `CHELSIO_INLINE_CRYPTO`?
> 
> Yes, it appears so, and `CONFIG_TLS` as well, which I assumed would already be set in the Talos kernel config, but appears not to be. I guess I'll have to fire up `menuconfig` for the first time in 10 years and do this properly. :) Give me a few minutes and I'll update the PR.

frezbo:
> > Is this needed? `CHELSIO_INLINE_CRYPTO`?
> 
> Yes, it appears so, and `CONFIG_TLS` as well, which I assumed would already be set in the Talos kernel config, but appears not to be. I guess I'll have to fire up `menuconfig` for the first time in 10 years and do this properly. :) Give me a few minutes and I'll update the PR.

CONFIG_TLS seems to do more than that? https://cateee.net/lkddb/web-lkddb/TLS.html like offload all TLS to kernel?

dhess:
> > > Is this needed? `CHELSIO_INLINE_CRYPTO`?
> > 
> > 
> > Yes, it appears so, and `CONFIG_TLS` as well, which I assumed would already be set in the Talos kernel config, but appears not to be. I guess I'll have to fire up `menuconfig` for the first time in 10 years and do this properly. :) Give me a few minutes and I'll update the PR.
> 
> CONFIG_TLS seems to do more than that? https://cateee.net/lkddb/web-lkddb/TLS.html like offload all TLS to kernel?

Agreed, it does not appear to be necessary, after all.

dhess:
I think the patch is correct now. I also enabled `CONFIG_CHELSIO_IPSEC_INLINE` as that might be useful for anyone using Cilium IPsec support, though personally I am not.

dhess:
In other distros, firmware for these NICs is normally bundled in a `firmware-linux` or `firmware-linux-nonfree` package. The Chelsio firmware is included in this tarball:

https://github.com/siderolabs/pkgs/blob/1401505a95eebb1ff9e2baac6239baf822b7576f/linux-firmware/pkg.yaml#L8

But I assume that the Talos installer and install ISO don't include all the firmware from that tarball, only a subset?

frezbo:
> In other distros, firmware for these NICs is normally bundled in a `firmware-linux` or `firmware-linux-nonfree` package. The Chelsio firmware is included in this tarball:
> 
> https://github.com/siderolabs/pkgs/blob/1401505a95eebb1ff9e2baac6239baf822b7576f/linux-firmware/pkg.yaml#L8
> 
> But I assume that the Talos installer and install ISO don't include all the firmware from that tarball, only a subset?

Talos doesn't include any firmware starting from 1.6, custom firmware is extracted and shipped as extensions: https://github.com/siderolabs/extensions/tree/main/firmware

frezbo:
I can't push the changes since the PR is from an org, could you re-create from a fork, or I can create a different PR

dhess:
We would like to keep our fork for now, so I guess just create a new PR? Sorry for the hassle and thanks for the help with this.

dhess:
I've confirmed that this patch is sufficient to bring up Chelsio NICs on bare metal using a Talos ISO.

Changes are also needed to https://github.com/siderolabs/talos/blob/main/hack/modules-amd64.txt (and the `arm64` equivalent). I'm happy to prepare a patch for that, as well, if you're satisfied with this PR.

frezbo:
> I've confirmed that this patch is sufficient to bring up Chelsio NICs on bare metal using a Talos ISO.
> 
> Changes are also needed to https://github.com/siderolabs/talos/blob/main/hack/modules-amd64.txt (and the `arm64` equivalent). I'm happy to prepare a patch for that, as well, if you're satisfied with this PR.

so firmware is not needed?

dhess:
Firmware is not required, no. The driver will look for firmware and complain if it doesn't find it, but then proceeds with the NIC's current firmware.

A firmware system extension would be useful, however, and I may try to put one together, if I have time to figure out how.

frezbo:
> Firmware is not required, no. The driver will look for firmware and complain if it doesn't find it, but then proceeds with the NIC's current firmware.
> 
> A firmware system extension would be useful, however, and I may try to put one together, if I have time to figure out how.

https://github.com/siderolabs/extensions/tree/main/firmware it's really easy, copy just the chelsio files

frezbo:
The modules will be added as an extension, sicnce it's over 2megs

PR #817, opened 2023-09-27
feat: add Solarflare SFC9000 support

USA-RedDragon:
I've got multiple machines with these cards that I'd like to use Talos with


USA-RedDragon:
Closing as https://www.talos.dev/v1.5/advanced/customizing-the-kernel/ manages my usecase

smira:
In fact, as these cards should be relatively popular, we'd rather include it as a module.

frezbo (approved):

PR #832, opened 2023-11-13
kernel: compile kernel headers as a module

USA-RedDragon:
This is for the usecase of utilizing eBPF programs to extend the kernel or to trace CPU usage. This compiles as a module so that an extension can be used in the event a user would like to utilize this.

As an example <https://github.com/iovisor/bcc> requires the `kheaders` module: https://github.com/iovisor/bcc/blob/1d659c7f3388c272979198c91d10fb9887f74f66/INSTALL.md?plain=1#L49-L50

Companion extensions commit: https://github.com/siderolabs/extensions/pull/262


frezbo:
We already compile kernel headers as BTF, so bpf programs already can use it

smira:
> We already compile kernel headers as BTF, so bpf programs already can use it

BTF is way better, and `bcc` already supports that

USA-RedDragon:
Fair, had to dig into the tech more to understand.

For anyone looking at this in the future, if you are trying to use the profiler at https://github.com/iovisor/bcc/blob/master/tools/profile.py, you probably want https://github.com/iovisor/bpftrace which has `libbpf` support and should work on Talos

PR #834, opened 2023-11-17
feat: lvm2 thin support

### LINKED ISSUE #833: Reboot loop with error 'WARNING: unrecognized segment type thin-pool'
Booting/Rebooting a cluster node with a disk that include thin-pool / thin logical volumes causes the boot process to fail, triggering a reboot loop.

**Error Message**
```
...
Failed to activate logical volumes, exit code 5
...
WARNING: unrecognized segment type thin-pool'
WARNING: unrecognized segment type thin
```
(I'm running this in proxmox so don't have a convenient way to copy/paste the full output)

**Environment Setup**
Cluster Version: "v1.6.0-alpha.0-59-g0bd1bdd74"
- Each node has 2 disks, one boot another for PV storage
- drbd and dm-thin-pool kernel modules in the configuration
```
machine:
  install:
    extensions:
      - image: "ghcr.io/siderolabs/drbd:9.2.5-v1.6.0-alpha.0-17-g0ba9f81"
  kernel:
    modules:
      - name: drbd
        parameters:
          - usermode_helper=disabled
      - name: drbd_transport_tcp
      - name: dm-thin-pool
```
- First boot and cluster initialization works fine, this is prior to /dev/sdb disk being initialized
- Install https://github.com/piraeusdatastore/piraeus-operator
- Configure piraeus to use drbd on /dev/sdb and provision as a thin-pool

As long as the system remains online, I can provision thin-pool volumes and everything works as expected. On reboot it the system fails to load.

**Possible Fix** 
From what I gather, this error is due to LVM2 of talos not being compiled with thin-pool support which is used during boot to mount the local disks.
https://listman.redhat.com/archives/linux-lvm/2013-July/022321.html

Setting https://github.com/siderolabs/pkgs/blob/main/lvm2/pkg.yaml#L26 to `--with-thin=internal` may fix the issue but I'm not exactly sure how this works if a user doesn't enable the dm-thin-pool module.

ndbrew:
Attempt to fix the reboot loop that happens when Talos encounters a disk that has lvm-thin volumes on it during boot.

Fixes #833

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,6 +1,6 @@
 # THIS FILE WAS AUTOMATICALLY GENERATED, PLEASE DO NOT EDIT.
 #
-# Generated on 2023-11-08T10:54:12Z by kres latest.
+# Generated on 2023-11-20T10:44:06Z by kres latest.
 
 name: default
 concurrency:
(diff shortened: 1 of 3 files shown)


smira (approved):
thank you!

PR #851, opened 2023-12-13
feat: make IOMMU useful on amd64 platform

yoctozepto:
On ARM platform, the relevant VFIO modules are enabled already. The amd64 platform was missing them as well as the IRQ_REMAP to support devices which need IRQ remapping in addition to DMA remapping.


yoctozepto:
/me splitting this into IRQ_REMAP and VFIO per request on Slack

yoctozepto:
split

PR #852, opened 2023-12-13
feat: enable IRQ remapping on amd64

yoctozepto:
This enables the IOMMU to do IRQ remapping besides DMA remapping which is required by some devices to function with IOMMU.

Note IOMMU is disabled by default.


smira (approved):
thank you!

PR #853, opened 2023-12-13
feat: enable VFIO also on amd64

yoctozepto:
On arm64 platform, the relevant VFIO configs are enabled already. This patch synchronizes both.


yoctozepto:
sorry, bad content...

PR #861, opened 2023-12-28
feat: enable NBD

dhess:
This is required by the Ceph CSI RBD plugin.


smira:
New module:

```
kernel/drivers/block/nbd.ko
```

Size: 

```
            │   │   ├── [4.0K]  block
            │   │   │   └── [115K]  nbd.ko
```

smira:
I can't push back to this PR, as it's an organization fork.

smira:
forked to #862

nberlee:
@dhess can you please elaborate what you are trying to solve? I've been using the Ceph CSI RBD plugin for years now with Talos, as RBD just works in the Talos Kernel.

Are you trying to use https://docs.ceph.com/en/latest/man/8/rbd-nbd by any chance?

dhess:
@nberlee In certain configurations, Rook does not reclaim blocks from deleted PVCs its its free list. As far as I understand it, there are a number of situations where this can occur:

* https://github.com/rook/rook/issues/3832
* https://github.com/rook/rook/issues/11562
* https://github.com/elastisys/compliantkubernetes-kubespray/issues/144

We noticed this when our free storage count didn't increase after deleting certain volumes. For us, setting the `discard` option didn't suffice. This may have had something to do with our using `Retain` claims, but I don't recall the details.

An iron-clad way to fix this is to run the CSI `ReclaimSpaceCronJob`: https://github.com/csi-addons/kubernetes-csi-addons/blob/v0.5.0/docs/reclaimspace.md#reclaimspacecronjob. Ceph CSI supports this job, but only if you have the `nbd` kernel module, because it runs from user space (otherwise, `krbd` is sufficient/preferable to `nbd` for Ceph). Without the `nbd` module, these jobs will fail, which is how we noticed that it's necessary.

PR #862, opened 2023-12-28
feat: enable NBD

smira:
This is required by the Ceph CSI RBD plugin.

Forked from #861


smira:
New module:

```
kernel/drivers/block/nbd.ko
```

Size: 

```
            │   │   ├── [4.0K]  block
            │   │   │   └── [115K]  nbd.ko
```

frezbo (approved):

PR #864, opened 2024-01-10
feat: add perf tool package

aarnaud:
Add perf linux tool from the same kernel version used by Talos.

Useful to diag/debug using a privileged containers that as full control of the node. 

```
    securityContext:
          allowPrivilegeEscalation: true
          privileged: true
    volumeMounts:
          - name: hostfs
            mountPath: /rootfs
...
hostIPC: true
hostPID: true
hostNetwork: true
volumes:
    - name: hostfs
      hostPath:
          path: /
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,6 +72,7 @@ TARGETS += linux-firmware
 TARGETS += lvm2
 TARGETS += musl
 TARGETS += openssl
+TARGETS += perf
 TARGETS += raspberrypi-firmware
 TARGETS += runc
 TARGETS += sd-boot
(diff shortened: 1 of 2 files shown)


frezbo:
Is the plan to make this as an extension? Would like to avoid pulling this into talos rootfs, this can be easily made as an extension

frezbo:
> You right, I didn't got yet the difference between pkgs and extensions. 

things in pkgs usually go into talos rootfs
extensions are optional

> Is extensions are design to be use only by imager to create boot assets ?

yes

PR #869, opened 2024-01-17
feat: applesmc-t2 sensors

aarnaud:
Allow to support fan control with t2fand on Mac mini 2018 late to avoid overheat.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,8 @@ COMMON_ARGS += --build-arg=SOURCE_DATE_EPOCH=$(SOURCE_DATE_EPOCH)
 
 # targets defines all the available targets
 
-TARGETS = base
+TARGETS = applesmc-t2-pkg
+TARGETS += base
 TARGETS += ca-certificates
 TARGETS += cni
 TARGETS += containerd
--- a/applesmc-t2/files/Makefile
+++ b/applesmc-t2/files/Makefile
@@ -1,0 +1,17 @@
+obj-$(CONFIG_SENSORS_APPLESMC) += applesmc.o
+
+KVERSION := $(KERNELRELEASE)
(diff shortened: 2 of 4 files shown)


aarnaud:
Hi @smira do you think i can be added as optional package ?

FYI: it's come from https://github.com/t2linux/linux-t2-patches the module is patch to fix fan.

smira:
The problem is supporting out-of-tree kernel modules, each one is a big pain, e.g. recent update to Linux 6.6 broke zfs and drbd. So if we take this module, we'd have to support it forever, which is a lot of work, and we have to make a decision in each particular case - how many users do we expect to benefit from it, etc.

aarnaud:
Good morning @smira, 
I understand, it's mostly used for Home Lab that is useful to experiment things before push stuffs in professional context. There is some people like me with refurbished/recycle Mac mini hardware. It's mostly to simplify the usage of Talos and Sidero. 

I'm agree to maintain this module, and if the build break, we can disable the module until I fix the build. We can also add a disclaimer on the module, that is a best effort. 

Is that would work for you ?

smira:
I would rather go the other way around, and I would like a way for you to add your own keys to the trusted Linux kerying, so that you can build, sign and load your own modules

andrewrynhard:
> I would rather go the other way around, and I would like a way for you to add your own keys to the trusted Linux kerying, so that you can build, sign and load your own modules

I agree with this. This is really the only way we will be able to handle all of the various drivers out there.

aarnaud:
yes, if there is a way to do that it will be perfect, if the process is well defined. 
I also have history where we had to patch drivers or test some proprietary drivers too.

aarnaud:
I'm closing this PR, I don't have apple device with `applesmc-t2` anymore to test

PR #871, opened 2024-01-22
fix: enable FUSION_SPI driver

smira:
It seems to be used by VMWare when virtualizing storage device as a SCSI disk.

Investigation by James Ringer from the community Slack.

```
udevadm info -a -n /dev/sda | grep -oP 'DRIVERS?=="\K[^"]+'
sd
mptspi
```


smira:
Note: when backporting to 1.6.x, make it `=y`.

smira:
New module:

```
├── message
│   └── fusion
│       └── [ 66K]  mptspi.ko
```

frezbo (approved):

frezbo:
> Note: when backporting to 1.6.x, make it `=y`.

is this just to avoid updateing `modules.txt` in talos ?

smira:
> > Note: when backporting to 1.6.x, make it `=y`.
> 
> is this just to avoid updateing `modules.txt` in talos ?

no, Talos 1.6 doesn't have it as modules (storage stuff).

jameshearttech:
I tested v1.6.3 ISO on my end with SCSI controller and LSI Logic Parallel bus sharing option, but `talosctl disks` returns nothing so I guess the fix did not work. It would be helpful if the `--dmesg` parameter worked with the `--insecure` parameter.

jameshearttech:
The machine I created from the v1.6.3 ISO shows the version as v1.6.1 in the Talos console.

frezbo:
> I tested v1.6.3 ISO on my end with SCSI controller and LSI Logic Parallel bus sharing option and confirmed the fix works as expected!

if you have Talos already installed iso would just boot that

Issue #870, opened 2024-01-21
raspberrypi_firmware_version is old


anguslees:
Pkgfile says: `raspberrypi_firmware_version: 1.20230405`, which lacks support for rPi 5, in particular.

It seems the raspberrypi/firmware repo has stopped tagging 'releases', so I think talos needs to switch to using some other versioning scheme for this dependency.  Or alternatively, find another source for these files.

I can build my own custom image, but I think rPi-5 has sufficient visibility to warrant fixing this for everyone.

smira:
@frezbo can you take a look please?

frezbo:
probably renovate didn't pick it up

PR #876, opened 2024-01-24
chore: switch to git ref for raspberrypi firmware

frezbo:
Switch to using Git Ref for Raspberry Pi firmware. The repository stopped tagging releases.

Fixes: #870

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -58,12 +58,6 @@
         },
         {
             "matchPackageNames": [
-                "raspberrypi/firmware"
-            ],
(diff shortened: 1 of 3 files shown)


utkuozdemir (approved):

smira:
@frezbo should we get it backported to 1.6 to get newer firmware there?

frezbo:
> @frezbo should we get it backported to 1.6 to get newer firmware there?

do we need to? since it only is going to fix Rpi5 and we don't officially support it

PR #884, opened 2024-02-02
Remove bash scripts

rothgar:
Remove scripts that call `/bin/bash` from being included with packages.


smira:
lgtm, but you should also make `conform` happy :) 

most of that is trivial commit message formatting, for the GPG you'd need to follow https://github.com/siderolabs/rfcs/blob/main/002b-key-management.md

rothgar:
Closed in favor of #889

PR #891, opened 2024-02-12
Add new kernel options

tpapagian:
This PR adds `CONFIG_SECURITY_PATH` and `CONFIG_BPF_LSM` config options in both `arm64` and `amd64` kernels.


smira (approved):
thankl you!

PR #899, opened 2024-02-18
feat: disable PCI busmastering on bridges during boot

### LINKED ISSUE #898: enable CONFIG_EFI_DISABLE_PCI_DMA to improve boot security from malicious PCI hardware
Enable `CONFIG_EFI_DISABLE_PCI_DMA` to improve boot security from malicious PCI hardware

From the author Matthew Garrett:

```
Add an option to disable the busmaster bit in the control register on
all PCI bridges before calling ExitBootServices() and passing control
to the runtime kernel. System firmware may configure the IOMMU to prevent
malicious PCI devices from being able to attack the OS via DMA. However,
since firmware can't guarantee that the OS is IOMMU-aware, it will tear
down IOMMU configuration when ExitBootServices() is called. This leaves
a window between where a hostile device could still cause damage before
Linux configures the IOMMU again.
```

A caution from the author:
```
This option may cause failures with some poorly behaved hardware and
should not be enabled without testing. The kernel commandline options
"efi=disable_early_pci_dma" or "efi=no_disable_early_pci_dma" may be
used to override the default. Note that PCI devices downstream from PCI
bridges are disconnected from their drivers first, using the UEFI
driver model API, so that DMA can be disabled safely at the bridge
level.
```

More info about [DMA attacks](https://en.wikipedia.org/wiki/DMA_attack)
Source of quotes: [lore](https://lore.kernel.org/linux-efi/20200103113953.9571-18-ardb@kernel.org/)

nberlee:
Enables `CONFIG_EFI_DISABLE_PCI_DMA` to improve boot security to protect from malicious PCI hardware.

Fixes #898 

Not sure where CONFIG_TOOLS_SUPPORT_RELR comes from, this was added after make kernel-olddefconfig


frezbo:
> Not sure where CONFIG_TOOLS_SUPPORT_RELR comes from, this was added after make kernel-olddefconfig

sorry, that's on me from https://github.com/siderolabs/pkgs/commit/65006ed198f31e97a77ddfded52043182f2c6e92 (must not forgot to run `olddefconfig`

frezbo (approved):

PR #905, opened 2024-02-28
feat: update releases

smira:
Linux: 6.6.18
ZFS: 2.2.3 (should fix the Linux 6.6 build failure)
iPXE: bump to next main


frezbo (approved):
seems there's also an update for rpi firmware and xz

smira:
> seems there's also an update for rpi firmware and xz

yes, that's true. I don't want to put too much here for now - I want to see if ZFS got fixed first.

smira:
> > seems there's also an update for rpi firmware and xz
> 
> yes, that's true. I don't want to put too much here for now - I want to see if ZFS got fixed first.

and it did build :dancers:

PR #906, opened 2024-02-28
feat: build NVMe target module

aarnaud:
Build NVMe target module to allow manually load to expose nvme over network.


smira:
@aarnaud this looks cool, do you have some notes on how you see that as a complete solution? system extension with kernel modules? how do you expose NVMe from Talos side?

aarnaud:
For example, it helps to be compatible with https://github.com/democratic-csi/democratic-csi?tab=readme-ov-file#zol-zfs-generic-nfs-zfs-generic-iscsi-zfs-generic-smb-zfs-generic-nvmeof

So the missing part is to deploy a pod that contains sshd and nvmetcli on the server that expose nvmeof so the controller can connect on it and expose disk over the network. 

This CSI can also use spdk but the setup is more complex and I'm not sure there spdk implementation can be used as is instead of nvmet

aarnaud:
there is also this:
https://github.com/kubernetes-csi/csi-driver-nvmf/blob/master/doc/setup_kernel_nvmf_target.md

aarnaud:
Thanks @smira, just to learn, is there a tools, option in talos-pkg to fix the dependency of config, like you did ? 
https://github.com/siderolabs/pkgs/compare/9ea461089e0bdf1b20a4324a3f62e4dc119986c2..e5c0c79b8d62448875bb21ca5701eacccd07c7e9

Also, is there plan to back-port this feature in 1.6 ?

PR #920, opened 2024-03-19
backport: feat: update dependencies for Talos 1.6.7

smira:
Linux: 6.1.82
Linux firmware: 20240312
DRBD: 9.2.8

Fixes https://github.com/siderolabs/extensions/issues/340


kernel/build/pkg.yaml
```diff
@@ -19,9 +19,6 @@ steps:
 
         make {{ .BUILD_ARG_KERNEL_TARGET }}
       {{ else }}
-      - |
```
frezbo, line 22: good call :sweat_smile:

PR #923, opened 2024-03-22
feat: configure kernel to include AppArmor LSM

konrader:
Enable AppArmor LSM in kernel.

https://www.kernel.org/doc/html/v6.1/admin-guide/LSM/apparmor.html


frezbo:
@konrader could you create the PR from a fork, I cannot push to a fork from an org. Or i can create a PR myself

konrader:
I didn't know that. I'll fix a personal fork and do the PR from there.

PR #926, opened 2024-03-27
chore: drop efi runtime services test framework

frezbo:
Drop the EFI runtime services test framework.

Ref: https://cateee.net/lkddb/web-lkddb/EFI_TEST.html


frezbo:
This was already disabled for amd64, let's make arm64 also similar.

Unix4ever (approved):

PR #934, opened 2024-04-05
feat: enable zboot on arm64 with zstd compression

smira:
Using zstd instead of xz for amd64, as xz lately can't be really trusted.


rsmitty (approved):

smira:
> [!NOTE]
> This is Talos 1.8+ change.

PR #936, opened 2024-04-09
feat: added support for official Quectel USB drivers

roeeklinger:


andrewrynhard:
We should put these under the `drivers` directory. Then I can kick off a build and see where things fail.

smira:
First of all, your PR is not against `main` - we don't support PRs to older release branches.

Second, we try to avoid 3rd-party Linux modules. Is support for it upstream in Linux kernel?

smira:
> Would that be unsupported?

yes, and the repository it's coming from looks like an attempt to do a supply chain attack?

roeeklinger:
I understand, thanks.

What do you mean? I just took the official drivers I got from Quectel and uploaded them to my own Github release repo for testing

smira:
Empty repository with some random files uploaded to releases.. no reproducibility, no source code, that looks suspicious.

No link to the source even. No way to update. Certainly no-go in the way it is right now.

andrewrynhard:
I failed to communicate that this is very much a WIP PR for @roeeklinger. They are evaluating Omni and need these drivers. @roeeklinger is running into issues with the build and asked for some help. I advised him to submit a PR here so we can help on a platform better suited for reviewing code.

@roeeklinger Third party drivers have made things difficult for us in the past so we are _very_ hesitant to include them. Third parties tend to move slow and the kernel moves fast. That then makes things break in our builds. Just some background here.

PR #946, opened 2024-04-25
feat: compress amd64 Linux kernel using zstd

smira:
This increases the kernel size, but moves us away from xz and improves decompression speed.

Also enable ZSTD firmware loading (not used yet).

Linux arm64 kernel is compressed with zstd already in 1.8+.


dsseng (approved):

kernel/build/config-amd64
```diff
@@ -16,6 +16,7 @@ CONFIG_CC_CAN_LINK_STATIC=y
 CONFIG_CC_HAS_ASM_GOTO_OUTPUT=y
 CONFIG_CC_HAS_ASM_GOTO_TIED_OUTPUT=y
 CONFIG_GCC_ASM_GOTO_OUTPUT_WORKAROUND=y
+CONFIG_TOOLS_SUPPORT_RELR=y
```
frezbo, line 19: this thing is a pain now, `menuconfig` removes them, `olddefconfig` adds it back :sweat_smile:

frezbo (commented):
does that mean we have to compress firmware with zstd?

frezbo (approved):

smira:
> does that mean we have to compress firmware with zstd?

I don't think we compress firmware at all, but it can load compressed if we decide to; I think having support doesn't hurt

frezbo:
> > does that mean we have to compress firmware with zstd?
> 
> I don't think we compress firmware at all, but it can load compressed if we decide to; I think having support doesn't hurt

cool

dsseng:
> > does that mean we have to compress firmware with zstd?
> 
> I don't think we compress firmware at all, but it can load compressed if we decide to; I think having support doesn't hurt

I can try looking into firmware compression someday. Perhaps we should open an issue for that to be looked into. Also we might want to consider further reducing firmware size by cutting out wireless and other firmwares we don't actually use because drivers aren't enabled

PR #971, opened 2024-06-05
feat: enable SCTP support

smira:
Fixes https://github.com/siderolabs/talos/issues/8871


DmitriyMV (approved):

frezbo:
@smira I wonder if we had disabled this due to some vulnerabilities, I can't recollect if it was SCTP for something else

PR #974, opened 2024-06-12
feat: enable CONFIG_KSM

stevefan1999-personal:
This is used in-conjunction with Kata Container, where each hypervisor would likely to have the same Kata kernel, exactly mapped page-to-page. We can use KSM here to have some memory back with little to no cost, maybe saving 20-30MB for each Cloud Hypervisor instance.


smira:
Note: `ksmd` should be explicitly enabled via the sysfs
(see https://www.kernel.org/doc/Documentation/vm/ksm.txt).

smira (approved):

PR #979, opened 2024-06-19
feat: add 'apparmor' package

konrader:
Include AppArmor userland tool /sbin/apparmor_parser in rootfs to enable AppArmor support in containerd

As discussed in https://github.com/siderolabs/talos/issues/8715

### DIFF
--- a/Pkgfile
+++ b/Pkgfile
@@ -173,5 +173,10 @@ vars:
   zfs_version: 2.2.4
   zfs_sha256: 9790905f7683d41759418e1ef3432828c31116654ff040e91356ff1c21c31ec0
   zfs_sha512: 1d17e30573d594fb5c9ea77cde104616dca362fed7530296816d1b55173594f66170fcfb23ab57c27074f85b79d3eb557b4ee9a1c420e507b2434a7902d8dcc1
+
+  # renovate: datasource=git-tags depName=https://gitlab.com/apparmor/apparmor.git
+  apparmor_version: v3.1.7
+  apparmor_sha256: 64494bd99fa6547a9cbdb4fc6bc732451a02dd19e6eb70eab977b239632151eb
+  apparmor_sha512: cfd6b0afb98d4559c16a6a2e23ca16ee9d86325fc6059313df5d3e8feba3d398f96a5754d3880dd2cafb2e7b1a06bd789d62cd36aaf993e46290f6311bb49dac
 labels:
   org.opencontainers.image.source: https://github.com/siderolabs/pkgs
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,6 +1,6 @@
 # THIS FILE WAS AUTOMATICALLY GENERATED, PLEASE DO NOT EDIT.
 #
-# Generated on 2024-05-28T12:19:20Z by kres a914cae.
+# Generated on 2024-06-25T05:52:46Z by kres 4c9f215.
 
 name: default
 concurrency:
@@ -33,7 +33,7 @@ jobs:
       labels: ${{ steps.retrieve-pr-labels.outputs.result }}
     services:
       buildkitd:
-        image: moby/buildkit:v0.13.2
+        image: moby/buildkit:v0.14.0
         options: --privileged
         ports:
           - 1234:1234
@@ -135,7 +135,7 @@ jobs:
       - default
     services:
       buildkitd:
-        image: moby/buildkit:v0.13.2
+        image: moby/buildkit:v0.14.0
         options: --privileged
         ports:
           - 1234:1234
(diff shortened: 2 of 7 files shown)


konrader:
This adds a 1MB userland tool `/sbin/apparmor_parser` in rootfs that `containerd` looks for as part of it's check for AppArmor support (https://github.com/containerd/containerd/blob/release/1.7/pkg/apparmor/apparmor_linux.go#L38)

It seem like apparmor LSM is not activated during Talos boot even if it's compiled into the kernel (after commit https://github.com/siderolabs/pkgs/commit/aefe000bd0cef8ec465b5fc14a1985ebba1fa429)
It can be activated with a kernel argument:
```
machine:
    install:
        extraKernelArgs:
            - security=apparmor
```
To verify that AppArmor LSM is enabled run `talosctl read /sys/kernel/security/lsm` that should return string that contains `apparmor` (amongst other LSM names)

konrader:
I've been testing Talos with the new `apparmor` package and if it's included in Talos it's actually convenient that AppArmor LSM isn't activated by default (even if it's already compiled into kernel) since existing Talos users won't get surprised by default AppArmor profile getting applied to their Pod containers on a Talos upgrade.

**Scenario 1**: Running vanilla Talos
AppArmor built into kernel: `yes`
apparmor pkg: `yes`
AppArmor LSM activated: `no`
**=> containerd won't apply default AppArmor profile to started containers** (same behaviour as before)

**Scenario 2**: Running Talos with config extraKernelArgs: `security=apparmor`
AppArmor built into kernel: `yes`
apparmor pkg: `yes`
AppArmor LSM activated: `yes`
**=> containerd will apply default AppArmor profile to started containers and user can start Pods and specify their own AppArmor profile via security context**

This way anyone could run a normal Talos release and enable AppArmor support via machine config without having to make a custom Talos build but at the same time anyone upgrading won't be surprised by containerd automatically starting to add default AppArmor profile to their deployed Pods.

Pkgfile
```diff
@@ -173,5 +173,10 @@ vars:
   zfs_version: 2.2.4
   zfs_sha256: 9790905f7683d41759418e1ef3432828c31116654ff040e91356ff1c21c31ec0
   zfs_sha512: 1d17e30573d594fb5c9ea77cde104616dca362fed7530296816d1b55173594f66170fcfb23ab57c27074f85b79d3eb557b4ee9a1c420e507b2434a7902d8dcc1
+
+  # renovate: datasource=git-refs versioning=git depName=https://gitlab.com/apparmor/apparmor.git
+  apparmor_version: v3.1.7
```
frezbo, line 178: there seems to be v4.0.1 available, any reason an older one was picked?

konrader: It seems like AppArmor 4.x is not "production ready" yet. When reading the release notes around 4.x they say both "release of the AppArmor that is in development" and "many of the features in apparmor 4 with require a development kernel" https://gitlab.com/apparmor/apparmor/-/releases/v4.0.1
Most Linux distributions that include AppArmor also package AppArmor 3.x so I was thinking of waiting bumping to AppArmor 4.1 until it comes out, which they indicate  "The following AppArmor 4.1 feature release is planned to be a regular release"

frezbo: makes sense :+1:

frezbo:
> I've been testing Talos with the new `apparmor` package and if it's included in Talos it's actually convenient that AppArmor LSM isn't activated by default (even if it's already compiled into kernel) since existing Talos users won't get surprised by default AppArmor profile getting applied to their Pod containers on a Talos upgrade.
> 
> **Scenario 1**: Running vanilla Talos AppArmor built into kernel: `yes` apparmor pkg: `yes` AppArmor LSM activated: `no` **=> containerd won't apply default AppArmor profile to started containers** (same behaviour as before)
> 
> **Scenario 2**: Running Talos with config extraKernelArgs: `security=apparmor` AppArmor built into kernel: `yes` apparmor pkg: `yes` AppArmor LSM activated: `yes` **=> containerd will apply default AppArmor profile to started containers and user can start Pods and specify their own AppArmor profile via security context**
> 
> This way anyone could run a normal Talos release and enable AppArmor support via machine config without having to make a custom Talos build but at the same time anyone upgrading won't be surprised by containerd automatically starting to add default AppArmor profile to their deployed Pods.

I think we can enable by default in 1.8, wdyt @smira ?

frezbo (approved):

frezbo:
Probably a neat addition to this: https://github.com/siderolabs/talos/pull/5975

smira:
> I think we can enable by default in 1.8, wdyt @smira ?

I feel we should document it _can_ be enabled for 1.8, and consider enabling it for 1.9 after we get some feedback, does it sound fair?

frezbo:
> > I think we can enable by default in 1.8, wdyt @smira ?
> 
> I feel we should document it _can_ be enabled for 1.8, and consider enabling it for 1.9 after we get some feedback, does it sound fair?

cool, I have added a release notes when i merged the talos PR, let's push to 1.9 then

PR #980, opened 2024-06-23
feat: enable CONFIG_X86_AMD_PSTATE

### LINKED ISSUE #978: Enable AMD pstate
Enable `CONFIG_X86_AMD_PSTATE`.

For any bare metal Talos system with an AMD CPU, this will allow better performance and power tuning.

See https://docs.kernel.org/admin-guide/pm/amd-pstate.html

jfroy:
> amd-pstate is the AMD CPU performance scaling driver that introduces a new CPU frequency control mechanism on modern AMD APU and CPU series in Linux kernel.

fixes #978


kernel/build/config-amd64
```diff
@@ -635,7 +635,8 @@ CONFIG_CPU_FREQ_GOV_SCHEDUTIL=y
 #
 CONFIG_X86_INTEL_PSTATE=y
 # CONFIG_X86_PCC_CPUFREQ is not set
-# CONFIG_X86_AMD_PSTATE is not set
+CONFIG_X86_AMD_PSTATE=y
+CONFIG_X86_AMD_PSTATE_DEFAULT_MODE=3
```
frezbo, line 639: not sure if we should set this to active or passive?

jfroy: `CONFIG_X86_AMD_PSTATE_DEFAULT_MODE=3` (passive) is the default because it most closely matches the performance of `acpi-cpufreq`, which is used if amd-pstate is disabled. It feels appropriate to keep that as the default and require the user to select active or guided (e.g. `amd-pstate-epp`) and configure it.

smira (approved):

jfroy:
Thanks for the approval. I rebased which broke the signature, sorry! I think your process is you re-commit, signing with a trusted key, so I should have left it alone.

DmitriyMV:
@jfroy no worries. Rebased and signed using my key. Will merge when PR is done.

PR #988, opened 2024-07-10
fix: update ca-certificates in pkgs

smira:
I did an update in tools, but forgot to update pkgs as well.


TimJones (approved):

frezbo (approved):

smira:
https://spaas.drbd.io/ is down (fails the build)

PR #994, opened 2024-07-28
fix(kernel): array-index-out-of-bounds error on bpf

samip5:
Fixes https://github.com/siderolabs/talos/issues/8780


smira:
Thank you for your PR, I need to push it back to re-sign the commits, but as your PR originates from the GitHub organization, I can't do that. So I will open new PR with the same commit.

nberlee:
@samip5 this looks like it was 1:1 taken from my repo (the markdown + bashcode). I was happy to contribute this back, but I am less happy that this happens without knowing that this is happening. As i may have a reason for it. 

Also @smira if I would have known this pr would have been accepted like this, I would have openened a pr as i mentioned my fix in the issue

smira:
@nberlee sorry, I didn't pay too much attention to the original work, but the pain point with UBSAN & Cilium seems relatively high compared to the cost of managing the patch.

I would certainly prefer the patch to be backported to the LTS kernel by kernel maintainers than us keeping the patch around.

smira:
@nberlee the only thing I can offer right now is to revert that commit if you feel that is the right way

nberlee:
Thanks for responding @smira , its ok. I am on holiday right now. Its fine as it is. You don't need to revert it.

samip5:
> @samip5 this looks like it was 1:1 taken from my repo (the markdown + bashcode). I was happy to contribute this back, but I am less happy that this happens without knowing that this is happening. As i may have a reason for it. 
> 

Sorry about not saying anything or asking about it beforehand, I was however thinking if everything of the email headers are indeed needed.

Issue #700, opened 2023-04-04
Enable Intel Management Engine Interface (MEI)


uhthomas:
It looks like the Intel Management Engine Interface (MEI) is required to use Intel Arc. The i915 firmware does not work as the HuC firmware will fail to load.

See:

https://github.com/jellyfin/jellyfin/issues/9588

https://github.com/uhthomas/pkgs/commit/6a83361b8e4facfe551657e49bc71fa114d8be0f

https://gitlab.freedesktop.org/drm/intel/-/issues/7732

Ornias1993:
@smira this should be moved to the official extentions repo, as it requires an extention. Which I think is a reasonable fit as an official extention.

smira:
Yes, sure, but the issue itself  is not actionable

Ornias1993:
> Yes, sure, but the issue itself is not actionable

In what sense is it not-actionable?

smira:
I for example have no idea what needs to be done, if it was a PR or detailed steps, it would be easier.

Ornias1993:
Agreed...
They are speaking in enabling kernel flags in the linked issues...

However...
Ive checked the Talos kernel flags and this isnt disabled, afaik it should be enabled when unset.

I'll continue doing some more research and see if I can get a more specific pointer to work from. My arc card just arrived to test!

e3b0c442:
Confirmed on the kernel config.

Additionally, I was able to see during kernel configuration that the flag `CONFIG_INTEL_MEI` can also be set as a module, so unless I'm mistaken, we should indeed be able to achieve this with an extension similar to the NVIDIA GPU extension by packaging the modules.

This is likely beyond my level of expertise, but I'm at least going to try and get this kernel built and boot assets created so I can test this on my own as a working solution.

PhilippWitzmann:
Thank you for investigating this. I've spent the whole day debugging my jellyfin installation and talos cluster after trying to use my new arc gpu, which ultimately led me here. 

Happy to help.

uhthomas:
Glad my foundational work was helpful, thanks for verifying it.

I wonder what the right thing to do here is? An Intel ME extension? Or should the i915 extension be updated to enable the Intel management engine?

e3b0c442:
I think a separate ME extension makes more sense:
- There are i915 use cases (DG1 and prior) that don't need ME
- ME is an additional attack vector and has a history of exploitable vulnerabilities, so those that don't absolutely need it may not want to enable it
[...]

PR #997, opened 2024-07-31
Add Intel management engine modules for Intel Arc support

e3b0c442:
This PR adds the necessary modules for Intel Management Engine along with the single non-module config necessary to enable Intel Arc (DG2) discrete GPU support as discussed in #700.

A corresponding PR will be made against siderolabs/extensions to add extensions to enable Arc.


smira (approved):
thank you!

PR #1004, opened 2024-08-08
eudev selinux

dsseng:
Hacky for now as we need to decide whether to build and add libselinux or work around it


frezbo:
I would say add libselinux if that would allow us to build eudev without patches, makes it easier to maintain

dsseng:
Alright, will add

libselinux/pkg.yaml
```diff
@@ -0,0 +1,28 @@
...
+        sha512: "{{ .libselinux_sha512 }}"
+    prepare:
+      - |
+        tar -xzf libselinux.tar.gz --strip-components=1
+
+        patch -p1 < /pkg/patches/musl-lstat.patch
+        patch -p1 < /pkg/patches/selabel-digest-uninit.patch
+    build:
+      - |
+        make -j $(nproc) FTS_LDLIBS="-lfts -L/toolchain/lib"
```
dsseng, line 22: needs `libfts.so.0`

pcre2/pkg.yaml
```diff
@@ -0,0 +1,32 @@
...
+      - |
+        tar -xjf pcre.tar.bz2 --strip-components=1
+        mkdir build
+        cd build
+        ../configure \
+              --prefix="/usr" \
+              --enable-unicode-properties \
+              --enable-pcre216 \
+              --enable-pcre232 \
+              --disable-static
```
dsseng, line 20: Dropped gzip and bzip2 grep support here

libselinux/patches/selabel-digest-uninit.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/utils/selabel_digest.c b/utils/selabel_digest.c
+index 47aad21f..c574d3fd 100644
+--- a/utils/selabel_digest.c
++++ b/utils/selabel_digest.c
+@@ -65,7 +65,7 @@ int main(int argc, char **argv)
+ 	size_t digest_len, i, num_specfiles;
+ 
+ 	char cmd_buf[4096];
+-	char *cmd_ptr;
++	char *cmd_ptr = NULL;
```
dsseng, line 10: To be upstreamed

frezbo: let's also add a link to the patch at the top of the file if it was from alpine or some mailing list

dsseng: Ah, not sure. Came up with those myself based on compiler errors :muscle: :brain: :)

If any of those has been posted already then will tag for sure

libselinux/patches/musl-lstat.patch
```diff
@@ -0,0 +1,22 @@
+diff --git a/src/selinux_restorecon.c b/src/selinux_restorecon.c
```
dsseng, line 1:
```suggestion
# https://bugs.gentoo.org/905711#c10
diff --git a/src/selinux_restorecon.c b/src/selinux_restorecon.c
```

libselinux/patches/selabel-digest-uninit.patch
dsseng: Could not google this up. Marked one found in Gentoo as well. It's not exactly the same since I made my myself from the current tree, but the idea is the same.

dsseng:
SELinux is now being implemented for systemd-udevd, and eudev is removed from 1.9. Closing as outdated

PR #1007, opened 2024-08-13
fix: add virtio-net GSO issue patch

smira:
Credits go to @frezbo for finding this patch.

See https://www.spinics.net/lists/stable/msg763970.html


frezbo (approved):

smira:
I verified locally that it fixes the problem.

PR #1025, opened 2024-09-10
fix: add mpt3sas UBSAN patches

smira:
Ref: https://github.com/torvalds/linux/commit/fd7090e384725edb1910a4b0a9c51007858f2c81

Ref: https://bugs.launchpad.net/ubuntu/+source/linux/+bug/2028830

Without these patches, the console is filled with UBSAN warnings from mpt3sas, and disk discovery seems to be wrong (?).


frezbo (approved):

smira:
Tested this with the bare-metal hardware with mpt3sas driver, works without issues now

bernardgut:
Nice thanks!

PR #1038, opened 2024-09-29
feat: replace eudev with systemd-udevd

dsseng:
lvm2 might need some attention as it's dependent on libudev


Pkgfile
```diff
@@ -113,6 +113,16 @@ vars:
...
+  # renovate: datasource=github-tags extractVersion=^(?<version>.*)$ depName=SELinuxProject/selinux
+  libselinux_version: 3.7
+  libselinux_sha256: ea03f42d13a4f95757997dba8cf0b26321fac5d2f164418b4cc856a92d2b17bd
+  libselinux_sha512: e949c20b606c50ad521b9592ce55ad6658e8c4b24d9838028f5aba0a4fc762b6d0d0d0d207f5bef7a2e41485e12d91382fa6090df27152dbb40071b273419352
+
+  # renovate: datasource=github-tags extractVersion=^(?<version>.*)$ depName=SELinuxProject/selinux
+  libsepol_version: 3.7
+  libsepol_sha256: cd741e25244e7ef6cd934d633614131a266c3eaeab33d8bfa45e8a93b45cc901
+  libsepol_sha512: 85d12d0ba5a7a3225f08d041a18fd59641608db5e0a78a1e9649754e45be54a807cd422d4889b88da6e806b4af546336c7a0913448f08ac33dc6ffb983890ef8
+
```
dsseng, line 125: Just as in tools repo we might want to set a common selinux_version for all the userspace SE parts

systemd-udevd/pkg.yaml
```diff
@@ -0,0 +1,114 @@
...
+    build:
+      - |
+        # rules.d/64-btrfs.rules is disabled in our build, other are static
+        # Buildable rules from rules.d/meson.build
+        ninja -j $(nproc) -C build udevadm rules.d/50-udev-default.rules rules.d/71-seat.rules rules.d/90-vconsole.rules \
+          rules.d/60-persistent-storage.rules rules.d/70-uaccess.rules rules.d/73-seat-late.rules rules.d/99-systemd.rules \
+          ata_id cdrom_id dmi_memory_id fido_id iocost mtd_probe scsi_id v4l_id \
+          libudev.so.1.7.9 libudev.a src/libudev/libudev.pc src/udev/udev.pc
+        # Build all instead and copy necessary parts?
+        # ninja -j $(nproc) -C build
```
dsseng, line 81: Here we either have to build 876 targets with a bulky command we need to update on some systemd updates (.so version change, new/removed udev rules) or build 1318 targets together, do ninja install and then copy directories we need into the image

systemd-udevd/pkg.yaml
```diff
@@ -0,0 +1,114 @@
...
+
+        # From src/udev/meson.build, used by rules
+        cp build/{ata_id,cdrom_id,dmi_memory_id,fido_id,iocost,mtd_probe,scsi_id,v4l_id} /rootfs/usr/lib/udev/
+
+        mkdir -p /rootfs/lib/pkgconfig
+        mkdir -p /rootfs/usr/share/pkgconfig
+        cp build/src/libudev/libudev.pc /rootfs/lib/pkgconfig/
+        cp build/src/udev/udev.pc /rootfs/usr/share/pkgconfig/
+
+        cp build/libudev.{a,so{.1.7.9,.1,}} /rootfs/lib/
```
dsseng, line 105: No `libudev.la` built here, maybe that is important for some builds

systemd-udevd/pkg.yaml
```diff
@@ -0,0 +1,114 @@
...
+          -Dopenssl=disabled \
+          -Dgshadow=false \
+          -Dutmp=false \
+          -Duserdb=false \
+          -Ddbus=disabled \
+          -Dglib=disabled \
+          -Dlink-udev-shared=false \
+          -Dman=disabled \
+          -Dtests=false \
+          -Dwerror=false
```
dsseng, line 71: We get some untrue warnings here. A compiler update might make those go away.

image
image

lvm2/pkg.yaml
```diff
@@ -4,7 +4,7 @@ shell: /toolchain/bin/bash
 dependencies:
   - stage: base
   - stage: libaio
-  - stage: eudev
+  - stage: systemd-udevd
```
dsseng, line 7:
```
4.868 configure: error: Package requirements (libudev >= 143) were not met:
4.868 
4.868 No package 'libudev' found
4.868 
4.868 Consider adjusting the PKG_CONFIG_PATH environment variable if you
4.868 installed software in a non-standard prefix.
4.868 
4.868 Alternatively, you may set the environment variables UDEV_CFLAGS
4.868 and UDEV_LIBS to avoid the need to call pkg-config.
4.868 See the pkg-config man page for more details.
```

systemd-udevd/pkg.yaml
```diff
@@ -0,0 +1,114 @@
...
+        mkdir -p /rootfs/usr/bin
+        cp build/udevadm /rootfs/usr/bin/
+        mkdir -p /rootfs/sbin
+        ln -s /rootfs/usr/bin/udevadm /rootfs/sbin/udevadm
+        ln -s /rootfs/usr/bin/udevadm /rootfs/sbin/systemd-udevd
+
+        mkdir -p /rootfs/usr/lib/udev/rules.d/
+        cp build/rules.d/*.rules /rootfs/usr/lib/udev/rules.d/
+        cp rules.d/*.rules /rootfs/usr/lib/udev/rules.d/
+        rm -f /rootfs/usr/lib/udev/rules.d/{60-cdrom_id.rules,60-persistent-alsa.rules,60-persistent-v4l.rules,64-btrfs.rules,70-joystick.rules,70-mouse.rules,70-touchpad.rules,78-sound-card.rules,80-net-name-slot.rules}
```
dsseng, line 93: We might want to remove more rules from our package.

Left: systemd-udevd, right: eudev:
image

lvm2/pkg.yaml
dsseng: Fixed, but something else regressed

dsseng: Yes, it's not just me: https://github.com/siderolabs/pkgs/actions/runs/11091800704/job/30816036407?pr=1038#step:9:5646

systemd-udevd/pkg.yaml
```diff
@@ -0,0 +1,114 @@
...
+
+        # From src/udev/meson.build, used by rules
+        cp build/{ata_id,cdrom_id,dmi_memory_id,fido_id,iocost,mtd_probe,scsi_id,v4l_id} /rootfs/usr/lib/udev/
+
+        mkdir -p /rootfs/usr/lib/pkgconfig
+        mkdir -p /rootfs/usr/share/pkgconfig
+        cp build/src/libudev/libudev.pc /rootfs/usr/lib/pkgconfig/
+        cp build/src/udev/udev.pc /rootfs/usr/share/pkgconfig/
+
+        cp build/libudev.{a,so{.1.7.9,.1,}} /rootfs/usr/lib/
```
dsseng, line 105: Perhaps we should create `libudev.so.1` and `libudev.so` as symlinks here. Will fix after major stuff

systemd-udevd/pkg.yaml
dsseng: reviewing the following extra rules: `60-dmi-id.rules 60-infiniband.rules 60-persistent-media-controller.rules 60-persistent-storage-mtd.rules 70-power-switch.rules 70-uaccess.rules 71-seat.rules 73-seat-late.rules 80-net-setup-link.rules 82-net-auto-link-local.rules 90-iocost.rules 90-vconsole.rules 99-systemd.rules`

dsseng: `99-systemd.rules` might be useless for us as it just points which targets should be activated. `71-seat.rules` might actually cause issues as we don't have logind. `90-vconsole.rules` also refers to systemd binaries and parts we have no intention to include.

lvm2/pkg.yaml
dsseng: Fixing the thin archive issue

systemd-udevd/pkg.yaml
```diff
@@ -0,0 +1,115 @@
...
+          libudev.so.1.7.9 libudev.a src/libudev/libudev.pc src/udev/udev.pc
+        # Build all instead and copy necessary parts?
+        # ninja -j $(nproc) -C build
+    install:
+      - |
+        mkdir -p /rootfs/usr/bin
+        cp build/udevadm /rootfs/usr/bin/
+        mkdir -p /rootfs/sbin
+        ln -s /rootfs/usr/bin/udevadm /rootfs/sbin/udevadm
+        ln -s /rootfs/usr/bin/udevadm /rootfs/sbin/systemd-udevd
```
dsseng, line 89: fixing bad symlinks

reproducibility/pkg.yaml
```diff
@@ -10,7 +10,7 @@ dependencies:
   - stage: containerd
   - stage: cryptsetup
   - stage: dosfstools
-  - stage: eudev
+  - stage: systemd-udevd
```
smira, line 13: can you please sort them?

systemd-udevd/pkg.yaml
smira: If that doesn't require more patches, I'd say build all and copy what we need (will be easier to maintain in the future).

Do systemd udev rules depend on something like `systemd-run`?

smira (approved):

lvm2/pkg.yaml
```diff
@@ -4,14 +4,16 @@ shell: /toolchain/bin/bash
...
   - stage: util-linux
 steps:
   - sources:
       - url: https://mirrors.kernel.org/sourceware/lvm2/LVM2.{{ .lvm2_version | replace "_" "." }}.tgz
         destination: lvm2.tar.gz
         sha256: "{{ .lvm2_sha256 }}"
         sha512: "{{ .lvm2_sha512 }}"
+    env:
+      LIBS: -l:libcap.a -L/toolchain/lib
```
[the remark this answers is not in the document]
dsseng: Libcap, I link it here statically. However I didn't test lvm2 (just did a qemu boot test and it got network working) so this might be wrong

systemd-udevd/pkg.yaml
dsseng: > Do systemd udev rules depend on something like systemd-run?

Looks like no, but I might need to proofread them anyway as we got a few more we are likely to not need (seat stuff, 99-systemd)

dsseng: Ok, here's the file tree of meson install results. https://tpaste.us/DrEo

Seems easier to maintain, and I'll try to minimize the components installed even more

systemd-udevd/pkg.yaml
dsseng: Current file list: https://tpaste.us/1yxz

systemd-udevd/pkg.yaml
```diff
@@ -0,0 +1,127 @@
...
+
+        mkdir -p /rootfs/sbin
+        ln -s /usr/bin/udevadm /rootfs/sbin/udevadm
+        ln -s /usr/bin/udevadm /rootfs/sbin/systemd-udevd
+
+        mkdir -p /rootfs/usr/lib/
+        cp -r /installroot/usr/lib/udev /rootfs/usr/lib/
+        cp --preserve=links /installroot/usr/lib/libudev.* /rootfs/usr/lib/
+        rm -f /rootfs/usr/lib/udev/hwdb.d/README
+        rm -f /rootfs/usr/lib/udev/rules.d/{README,60-cdrom_id.rules,60-persistent-alsa.rules,60-persistent-v4l.rules,64-btrfs.rules,70-joystick.rules,70-mouse.rules,70-touchpad.rules,78-sound-card.rules,80-net-name-slot.rules,90-vconsole.rules,99-systemd.rules}
```
dsseng, line 113: Should we keep [90-iocost.rules](https://github.com/systemd/systemd/blob/d713c131a58b4c2b46ff0edc0fea26037edb2140/rules.d/90-iocost.rules) (untested) and should we fix hwdb or drop it? [hwdb errors](https://tpaste.us/W5YN) due to missing hwdb binary file which we have to build using systemd-hwdb

dsseng, line 113: Should we keep [90-iocost.rules](https://github.com/systemd/systemd/blob/d713c131a58b4c2b46ff0edc0fea26037edb2140/rules.d/90-iocost.rules) (untested) and should we fix hwdb or drop it? [hwdb errors](https://tpaste.us/W5YN) due to missing hwdb binary file which we have to build using systemd-hwdb

lvm2/pkg.yaml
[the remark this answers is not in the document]
smira: `-L` is global here afaik, it's not scoped to a single `-l`, if we build `libcap`, we can link dynamically as well?

systemd-udevd/pkg.yaml
smira: I believe `hwdb` is required, otherwise it won't do proper stuff? I thought we had it with eudev

lvm2/pkg.yaml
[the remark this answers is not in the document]
smira: I'm confused as we build `libcap` here in pkgs as well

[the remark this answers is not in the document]
dsseng: I added it as it's a dynamic dep of udev. Will switch (or try to switch) other packages to dynamic libcap

systemd-udevd/pkg.yaml
dsseng: We didn't, see my other comment where I looked at the difference in package contents. Anyway looking into that rn to see what it does except for HID devices and if we need to fix it or disable (via systemd build opts). iocost is also interesting, somewhat relevant to containers as per first Google search. If it can be useful to us then hwdb is apparently a must for it

dsseng: Okay, iocost fields are not populated in the default rules. However [it might be useful](https://tangchq74.github.io/IOCost_IEEE_TopPicks_23.pdf) and unless `IOCOST_SOLUTIONS` is set for a disk device it does nothing

dsseng:
Should we split this up into adding SELinux libs + replacing udev as 2 separate commits?

smira:
> Should we split this up into adding SELinux libs + replacing udev as 2 separate commits?

These can be separate PRs, but not commits.

dsseng:
> > Should we split this up into adding SELinux libs + replacing udev as 2 separate commits?
> 
> These can be separate PRs, but not commits.

Well, yes, and PRs as well. Will submit SELinux libraries separately once approved

dsseng:
SELinux support appears broken, once I patch libselinux to say SELinux is enabled despite we don't store a config on rootfs udev crashes. Anyway that's a separate task

jfroy:
This is nice, but maintaining that patch set is going to be pain. I wonder if the glibc extension should be brought into pkgs and just accept that a barebones glibc will exists on Talos.

smira:
> This is nice, but maintaining that patch set is going to be pain. I wonder if the glibc extension should be brought into pkgs and just accept that a barebones glibc will exists on Talos.

Yes, that would be easier, but this goes against the security concepts. 

If you use NVIDIA container runtime, you certainly accept the security risks, as it's usually very bad compared to bare `runc`, and in that case having `glibc` is fine as well, as it might not be the worst thing. But having `glibc` in every Talos install does't sound like a great idea.

PR #1039, opened 2024-09-30
feat: add libselinux, libsepol, pcre2 and libcap

dsseng:


Makefile
```diff
@@ -60,17 +60,19 @@ TARGETS += iptables
 TARGETS += ipxe
 TARGETS += kmod
 TARGETS += libaio
-TARGETS += libattr
```
smira, line 63: this looks like a bug in previous commits, but this should stay

smira (approved):

Makefile
dsseng: I did a make rekres. Should I just add `TARGETS += libattr` back?

dsseng:
@smira please also check libcap, decided to add it here as well

PR #1042, opened 2024-10-02
fix: drop the LVM2 udev lvm rule

smira:
This rule depends on `systemd-run`, and it won't work in Talos anyways, but it conflicts with Talos activating LVM volumes, as the call to `pvscan` updates the cache as if the volume got already activated, while in fact it hasn't been activated yet.

### DIFF
--- a/lvm2/pkg.yaml
+++ b/lvm2/pkg.yaml
@@ -99,6 +99,9 @@ steps:
     install:
       - |
         make DESTDIR=/rootfs install
+
+        # LVM activation is handled by Talos itself
+        rm -f /rootfs/usr/lib/udev/rules.d/69-dm-lvm.rules
 finalize:
   - from: /rootfs
     to: /


lvm2/pkg.yaml
```diff
@@ -99,6 +99,9 @@ steps:
     install:
       - |
         make DESTDIR=/rootfs install
+
+        # LVM activation is handled by Talos itself
+        rm -f /rootfs/usr/lib/udev/rules.d/69-dm-lvm.rules
```
dsseng, line 104: Isn't https://github.com/lvmteam/lvm2/blob/main/udev/95-dm-notify.rules.in also unnecessary? We seem not to have dmsetup in our rootfs if I'm not mistaken

dsseng (dismissed):

lvm2/pkg.yaml
smira: probably, but I don't want to make bigger changes, as this is going to be backported to 1.8 - so the minimal changes required

dsseng (approved):

PR #1044, opened 2024-10-02
feat: add lpfc kernel module

Hexoplon:
Add lpfc kernel module, alongside required crct10dif module. This driver is used with Emulex Fibre Channel HBA adapters.

These changes are intended for use with a driver extension for use with the image factory.

[Issue on extension repo](https://github.com/siderolabs/talos/issues/9437)

PR is currently draft, as I am unsure if there is anything else that is required to change


smira:
I did some small fixes, this will be backported to 1.8.1

smira (approved):

PR #1047, opened 2024-10-04
fix: systemd-udevd: search for config in /usr/etc

dsseng:
This was breaking mdadm system extension. Added a patch to make udev honor the compile-time setting

Signed-off-by: Dmitry Sharshakov <dmitry.sharshakov@siderolabs.com>


dsseng:
Related: #9423

smira (approved):

PR #1048, opened 2024-10-05
feat: fix Mellanox Build & add missing Kernel Flags

feniarus:
This PR is for fixing the Mellanox  OFED Driver Package and adding the missing Kernel Flags for it to build.
The NVME Part of the Dirver is disabled because it has Source Code related problems.


mellanox-ofed/pkg.yaml
```diff
@@ -1,5 +1,5 @@
 name: mellanox-ofed-pkg
-variant: scratch
+variant: alpine
```
dsseng, line 2: Not sure whether or not we should do so. Why doesn't it work without `variant: alpine`, what's missing in our tools, and why not add it there?

dsseng:
> The NVME Part of the Dirver is disabled because it has Source Code related problems

Which precisely?

Also, your commit doesn't contain a second body line (well, just add description there) and is missing a signed-off-by line, which indicates accepting Developer Certificate of Origin. Please add it using git commit --amend -s

mellanox-ofed/pkg.yaml
```diff
@@ -13,29 +13,27 @@ steps:
       ARCH: {{ if eq .ARCH "aarch64"}}arm64{{ else if eq .ARCH "x86_64" }}x86_64{{ else }}unsupported{{ end }}
     prepare:
       - |
-        tar -xzf mellanox_ofed.tgz --strip-components=1
+        rm /bin/mktemp
+        ln -s /toolchain/bin/mktemp /bin/
```
dsseng, line 17: Could you please add a comment why is this replacement done?

Pkgfile
```diff
@@ -140,9 +140,9 @@ vars:
   lvm2_sha256: 4c5a6923bd1ace7ce04474608a84937ce053ba91b1ace9f0b0017268e732dc7c
   lvm2_sha512: 17cd24ceee8026481566824b688dafd03ec816201d5cb3549cb7fc8a36f4cdaa982faaef4dcd26debfe775dea5ffa2744798164314ea6dc99a84f8ccccfc33ff
 
-  mellanox_ofed_version: 5.9-0.5.6.0
-  mellanox_ofed_sha256: 4503258cbe92b00c734e612c3a7ad1d71e023fdffae2a2c119f7b537505e499c
-  mellanox_ofed_sha512: 58604ea89aa8351727532c48f0c59b4e533ba8bfcef9533f45d94e15ffdcf3a5c464398706cad14ebf3826b132972bd044fadbf7f047e60bdb0c2a454c96acd7
+  mellanox_ofed_version: 24.07-0.6.1.0
```
dsseng, line 143: Please also include a note about the version update in the commit message

dsseng (commented):
Thank you for contributing! Could you please elaborate a bit on some parts unclear to me?

Makefile
```diff
@@ -84,6 +84,7 @@ TARGETS += xfsprogs
 TARGETS += kernel
 TARGETS += drbd-pkg
 TARGETS += gasket-driver-pkg
+TARGETS += mellanox-ofed-pkg
```
smira, line 87: this file is generated, instead `.kres.yaml` should be updated with `make rekres` after that

feniarus:
> > The NVME Part of the Dirver is disabled because it has Source Code related problems
> 
> Which precisely?
> 
> Also, your commit doesn't contain a second body line (well, just add description there) and is missing a signed-off-by line, which indicates accepting Developer Certificate of Origin. Please add it using git commit --amend -s

When trying to build with full NVME support, the build fails with the following error:

image

frezbo:
> This PR is for fixing the Mellanox OFED Driver Package and adding the missing Kernel Flags for it to build. The NVME Part of the Dirver is disabled because it has Source Code related problems.
> 
> Signed-off-by: Feniarus [101282511+feniarus@users.noreply.github.com](mailto:101282511+feniarus@users.noreply.github.com)

does this means it's better than the in-tree kernel mlx5 drivers, just trying to gauge if we need to build and maintain this? :thinking:

frezbo:
also wants to know if this co-exist with in-tree kernel drivers, might need to denylist udev loading mlx5 by default I guess

PR #1052, opened 2024-10-07
feat: enable CONFIG_XFRM_STATISTICS

PhilipSchmid:
For context, this config enables a /proc file with a bunch of error counters. Debugging information from the kernel is very sparse when you hit an issue with IPsec (packet corruption, incorrect key, etc.) and that file is often all you get.
Debugging without this information is really tough. And the overhead is expected to be negligible (typically once counter increment per packet) compared to the cost of encryption/decryption.


smira (approved):
thank you!

PR #1057, opened 2024-10-21
fix: default IOMMU mode to 'lazy'

smira:
Fixes https://github.com/siderolabs/talos/issues/9531

See also e.g. https://bugs.launchpad.net/ubuntu/+source/linux/+bug/2071471

This makes the default value better for IO performance, previous behavior can be restored with `iommu.strict=1`.


frezbo (approved):

dsseng (approved):
Shouldn't really reduce security, good. Anyway can be overriden

smira:
That's why we had it enforced:

```
CONFIG_IOMMU_DEFAULT_DMA_STRICT              |      y      |   kspp   |  self_protection   
```

PR #1065, opened 2024-10-25
fix: kernel netfilter v6 --set-mark in 6.6.58

nberlee:
When running kernel 6.6.58 without this patch CNIs that use ip6tables break because of a typo in the netfilter code, introduced in 6.6.57.

Cilium with ipv6 enabled for example outputs:
```
level=error msg="iptables rules full reconciliation failed, will retry another one later" error="failed to install rules: cannot install static proxy rules: unable to run 'ip6tables -t mangle -A CILIUM_PRE_mangle -m socket --transparent ! -o lo -m mark ! --mark 0x00000e00/0x00000f00 -m mark ! --mark 0x00000800/0x00000f00 -m comment --comment cilium: any->pod redirect proxied traffic to host proxy -j MARK --set-mark 0x00000200' iptables command: exit status 4 stderr=\"Warning: Extension MARK revision 0 not supported, missing kernel module?\\nip6tables v1.8.8 (nf_tables):  RULE_APPEND failed (No such file or directory): rule in chain CILIUM_PRE_mangle\\n\"" subsys=iptables
```


smira:
Sorry, Nico, I missed your PR while creating mine (#1066) - the patch is the same in the end. I'll put it for 1.8.x only, as in `main` we should get new kernel this week.

PR #1068, opened 2024-11-03
fix: enable nvme and 2.5gbit ethernet on nanopi-r5s

RemkoMolier:
Enable some ROCKCHIP related kernel options to allow for pcie to work.

This enables both the nvme as well as the other ethernet ports on the nanopi-r5s.


frezbo:
could you re-open against main, probably no need to go into release-1.8

RemkoMolier:
> could you re-open against main, probably no need to go into release-1.8

You are probably right, I'll do so this afternoon. Would there be a chance to get this backported to 1.8 anyway, so people can start using the nanopi-r5s from the 1.8 release in sbc-rockchip with its full functionality?

frezbo:
> > could you re-open against main, probably no need to go into release-1.8
> 
> You are probably right, I'll do so this afternoon. Would there be a chance to get this backported to 1.8 anyway, so people can start using the nanopi-r5s from the 1.8 release in sbc-rockchip with its full functionality?

will see if that makes sense to backport, usually only fixes go to backport

smira:
@RemkoMolier please rebase on top of `main`, we won't accept direct PRs to `release-1.8` anyways. This can be backported I guess.

RemkoMolier:
@smira thank you, will do so in a couple of hours.

RemkoMolier:
Reopened on main as #1072

PR #1075, opened 2024-11-05
feat: enable CONFIG_DM_CACHE

chifu1234:
https://github.com/siderolabs/pkgs/pull/1075


smira:
probably needs `DM_CACHE_SMQ` and both should be modules

smira (approved):

PR #1077, opened 2024-11-06
feat: update gcc to 14.2

smira:
Add some small fixes to make it compiple.


smira:
Bumped NVIDIA drivers, applied patches to make them compile with gcc 14.2.

Issue #990, opened 2024-07-13
FR: Build kernel variant with DVB subsystem enabled


samip5:
It seems that the kernel config being used, is disabling all of dvb subsystem. This is not great for tuner cards.
Thus I'm requesting it to be enabled.

My tuner card is https://www.linuxtv.org/wiki/index.php/Hauppauge_WinTV-quadHD_(DVB-T/T2/C) which then uses cx23885 driver in kernel. This driver doesn't seem to be included nor is the dvb subsystem as a whole.

smira:
This won't be included by default in the Talos kernel, as it's certainly a very niche requirement.

But you can prepare a PR with enabling required stuff as kernel modules, and push another to repackage as extensions. For example, https://github.com/siderolabs/extensions/tree/main/drivers/v4l-uvc

samip5:
The patchset for the required things to enable for dvb is quite large, even when I tried to focus it on just my card.
Related: https://github.com/skyssolutions/talos-kernel/commit/598d30e1d2151cea062790736ebf4a12c41f302a

samip5:
Fixed by #1102

PR #1089, opened 2024-11-18
feat: build dvb kernel modules and CX23885

samip5:
Fixes https://github.com/siderolabs/pkgs/issues/990
This also needs dvb firmware for said tuner, but this is needed first and these need to be packaged to an extension.


frezbo:
@samip5 are these also required for arm64? Also do you have a corresponding extensions PR? Otherwise this'll go into 1.10

samip5:
I'm not sure if the drivers will work on arm64, but I at least don't need them in arm64, so that's why I have added it only to amd64. I have a corresponding extensions pr for the firmware part, but not for the actual dvb modules yet.

So are you referring to the extension of including the kernel modules this builds as a extension?

frezbo:
> So are you referring to the extension of including the kernel modules this builds as a extension?

yes, these won;t be shipped with core talos

frezbo:
Re-created as #1102, since cannot push PR created from org

PR #1085, opened 2024-11-18
feat: update Linux to 6.12

smira:
Totally WIP, no third-party modules can be built.

PR #1090, opened 2024-11-19
Fix kernel URL substitution

attilaolah:
For kernels without a patch version (e.g. 6.12), the directory seems to be 6.x itnheai of 6.12.x.


kernel/prepare/pkg.yaml
```diff
@@ -5,7 +5,7 @@ dependencies:
   - image: '{{ .TOOLS_IMAGE }}'
 steps:
   - sources:
-      - url: https://cdn.kernel.org/pub/linux/kernel/v{{ regexReplaceAll ".\\d+\\.\\d+$" .linux_version "${1}" }}.x/linux-{{ .linux_version }}.tar.xz
+      - url: https://cdn.kernel.org/pub/linux/kernel/v{{ regexReplaceAll ".\\d+(\\.\\d+)?$" .linux_version "${1}" }}.x/linux-{{ .linux_version }}.tar.xz
```
attilaolah, line 8: Or we could simplify the regex to (\\.\\d+)+, that would leave only the major version, should also work, right?

attilaolah:
Oh I guess this can be closed in favour of #1085, which changes the URL to GitHub…

PR #1091, opened 2024-11-20
feat: add e2fsprogs

frezbo:
Add e2fsprogs.

Part-of: https://github.com/siderolabs/talos/issues/9746


frezbo:
Also verified that the build is reproducible locally.

smira (approved):

PR #1095, opened 2024-11-27
feat: update systemd to 256.8, fix cpuset/cgroupsv1

smira:
The cpuset/cgroupsv1 is required for cgroupsv1 support.

Fix systemd renovate rule.


frezbo (approved):

frezbo:
should we drop cgroupsv1 in 1.9?

smira:
> should we drop cgroupsv1 in 1.9?

the goal is 1.10 - we deprecate in 1.9

PR #1102, opened 2024-12-02
feat: build dvb kernel modules and CX23885

frezbo:
This is required to be able to use Hauppage WinTV-quadHD PCIe tuner card


samip5:
Related: https://github.com/siderolabs/extensions/pull/538

PR #1103, opened 2024-12-02
chore: bring in KSPP recommendations

frezbo:
Bring in new KSPP recommendations.

Ref: https://github.com/a13xp0p0v/kernel-hardening-checker/commit/12eb32d21203bd164271281c6266f161a300c40c


kernel/build/config-arm64
```diff
@@ -1122,7 +1122,7 @@ CONFIG_PCP_BATCH_SCALE_MAX=5
 CONFIG_PHYS_ADDR_T_64BIT=y
 CONFIG_MMU_NOTIFIER=y
 CONFIG_KSM=y
-CONFIG_DEFAULT_MMAP_MIN_ADDR=32768
+CONFIG_DEFAULT_MMAP_MIN_ADDR=65536
```
smira, line 1125: this seems like a bug in the checker script: https://cateee.net/lkddb/web-lkddb/DEFAULT_MMAP_MIN_ADDR.html

> On arm and other archs it should not be higher than 32768

smira: ah... `arm64/arm/` :)

smira (approved):

PR #1121, opened 2024-12-17
feat: update Linux firmware to 20241210

smira:
Latest release.


smira:
See https://github.com/siderolabs/extensions/issues/558

PR #1136, opened 2025-01-10
fix: adjust kernel options around ACPI/PCI/EFI

smira:
Align options borrowing from other distribution kernel configurations.


kernel/build/config-amd64
```diff
@@ -611,28 +612,34 @@ CONFIG_ACPI_THERMAL=y
...
-# CONFIG_ACPI_PCI_SLOT is not set
+CONFIG_ACPI_PCI_SLOT=y
 CONFIG_ACPI_CONTAINER=y
 CONFIG_ACPI_HOTPLUG_MEMORY=y
 CONFIG_ACPI_HOTPLUG_IOAPIC=y
 # CONFIG_ACPI_SBS is not set
-# CONFIG_ACPI_HED is not set
-# CONFIG_ACPI_BGRT is not set
+CONFIG_ACPI_HED=y
+CONFIG_ACPI_BGRT=y
```
dsseng, line 621: Do we really need this? Iirc that's a table supplying UEFI vendor Logo for OS boot screen

frezbo (approved):
does this give us any benefits?

kernel/build/config-amd64
smira: I can disable that as well, I thought it might be cool eventually to support showing this :)

smira:
> does this give us any benefits?

I don't have a clear regression to compare against, so this more of `kconfig-diff` and looking up each option.

kernel/build/config-amd64
dsseng: well, yes, ACPI table parsing shouldn't be heavy so okay

kernel/build/config-amd64
```diff
@@ -2083,6 +2095,9 @@ CONFIG_SYSFB=y
 # EFI (Extensible Firmware Interface) Support
 #
 CONFIG_EFI_ESRT=y
+CONFIG_EFI_VARS_PSTORE=y
```
dsseng, line 2098: Is this disabled by default? Polluting UEFI NVRAM on crashes by default might not be desirable

dsseng (approved):

kernel/build/config-amd64
frezbo: well i don;t think it hurts to show the vendor logo

kernel/build/config-amd64
smira: good point, I think we should disable it

smira: removed

PR #1147, opened 2025-01-27
feat: update Linux to 6.12.11

smira:
Latest Linux LTS 6.12.x.


dsseng (approved):

kernel/build/config-arm64
```diff
@@ -35,16 +35,8 @@ CONFIG_LOCALVERSION="-talos"
 # CONFIG_LOCALVERSION_AUTO is not set
 CONFIG_BUILD_SALT=""
 CONFIG_HAVE_KERNEL_GZIP=y
-CONFIG_HAVE_KERNEL_LZMA=y
```
frezbo, line 38: would this break some upragde where initrd was xz compressed, I guess it shouldn;t matter

frezbo (approved):

kernel/build/config-arm64
smira: it looks like they cleaned up some options, it is kernel compression, and we enabled ZSTD explicitly.

Issue #1151, opened 2025-02-03
Enable usrmerge and disable multilib universally


dsseng:
As Talos only uses 64-bit executables, it makes sense for `/lib`, `/lib64`, `/usr/lib64` to all be symlinks pointing to `/usr/lib` as the unified directory for libraries. Similarly, `/bin` points to `/usr/bin` and `/sbin` -> `/usr/sbin`.

During the hermetic build epic I made sure toolchain and tools all follow these filesystem hierarchy standards, but to avoid too many breaking changes this should be done in pkgs only after other refactors (incl. for example **deduplication of packages like musl**)

PR #1152, opened 2025-02-04
feat: hermetic pkgs build

dsseng:
feat: bootstrapped build

Build pkgs using only [Stageˣ]-derived tools image, without relying on Alpine

Fixes: siderolabs/toolchain#129
Fixes: #1151

- feat: update bldr
- fix: remove /toolchain prefix
- feat: build what is possible using network: none
- feat: usrmerge and unified /usr/lib
- chore: use StageX bash instead of Alpine bash
- fix: make iptables build
- chore: move kernel modules under /usr/lib/modules
- refactor: remove git requirement
- test: check nvidia nonfree drivers' signatures


iptables/pkg.yaml
```diff
@@ -15,18 +15,20 @@ steps:
       - |
         tar -xf iptables.tar.xz --strip-components=1
 
+        patch -p1 < /pkg/patches/0001-fix-redefinition-on-musl.patch
```
dsseng, line 18: Weird how it broke. May be somehow related to relocation of headers

kernel/prepare/pkg.yaml
```diff
@@ -11,42 +11,20 @@ steps:
...
-        ln -s /toolchain/bin/awk /usr/bin/awk
-
-        mkdir -p /usr/lib/pkgconfig
-        ln -s /toolchain/include /usr/include
-
         make mrproper
       - |
-        cd /toolchain && git clone https://github.com/a13xp0p0v/kernel-hardening-checker.git
+        # TODO: remove
+        mkdir /toolchain && cd /toolchain && git clone https://github.com/a13xp0p0v/kernel-hardening-checker.git
```
dsseng, line 23: Why do we pull this with git and not just as a tar?

smira: No good reason, we want the latest & greatest, and it doesn't go into the build results, it's a linter.

dsseng: `wget https://github.com/a13xp0p0v/kernel-hardening-checker/archive/refs/heads/master.tar.gz` seems to work. Then git can perhaps be removed from tools

dsseng: Done using pulling from master. However I checked and this tool runs in the build step (before artifact creation). So if it's not pinned and gets replaced by malware it could affect the build. Also makes building older versions not guaranteed to succeed, because if some rule is added to KSPP and someone wants to build an old version, it won't let it work. Anyway, keeping mostly the same but using wget now

nonfree/kmod-nvidia/lts/pkg.yaml
```diff
@@ -32,12 +32,12 @@ steps:
...
-        cp /src/modules.builtin /rootfs/lib/modules/$(cat /src/include/config/kernel.release)/
-        cp /src/modules.builtin.modinfo /rootfs/lib/modules/$(cat /src/include/config/kernel.release)/
+        mkdir -p /rootfs/usr/lib/modules/$(cat /src/include/config/kernel.release)/
+        cp /src/modules.order /rootfs/usr/lib/modules/$(cat /src/include/config/kernel.release)/
+        cp /src/modules.builtin /rootfs/usr/lib/modules/$(cat /src/include/config/kernel.release)/
+        cp /src/modules.builtin.modinfo /rootfs/usr/lib/modules/$(cat /src/include/config/kernel.release)/
 
-        make -j $(nproc) modules_install SYSSRC=/src DEPMOD=/toolchain/bin/depmod INSTALL_MOD_PATH=/rootfs INSTALL_MOD_DIR=extras INSTALL_MOD_STRIP=1
+        make -j $(nproc) modules_install SYSSRC=/src INSTALL_MOD_PATH=/rootfs/usr INSTALL_MOD_DIR=extras INSTALL_MOD_STRIP=1
 finalize:
```
dsseng, line 41: Any reason we do not do this test on NVIDIA proprietary?
```suggestion
    test:
      - |
        # https://www.kernel.org/doc/html/v4.15/admin-guide/module-signing.html#signed-modules-and-stripping
        find /rootfs/usr/lib/modules -name '*.ko' -exec grep -FL '~Module signature appended~' {} \+
finalize:
```

nonfree/kmod-nvidia/production/pkg.yaml
```diff
@@ -32,12 +32,12 @@ steps:
...
-        cp /src/modules.builtin /rootfs/lib/modules/$(cat /src/include/config/kernel.release)/
-        cp /src/modules.builtin.modinfo /rootfs/lib/modules/$(cat /src/include/config/kernel.release)/
+        mkdir -p /rootfs/usr/lib/modules/$(cat /src/include/config/kernel.release)/
+        cp /src/modules.order /rootfs/usr/lib/modules/$(cat /src/include/config/kernel.release)/
+        cp /src/modules.builtin /rootfs/usr/lib/modules/$(cat /src/include/config/kernel.release)/
+        cp /src/modules.builtin.modinfo /rootfs/usr/lib/modules/$(cat /src/include/config/kernel.release)/
 
-        make -j $(nproc) modules_install SYSSRC=/src DEPMOD=/toolchain/bin/depmod INSTALL_MOD_PATH=/rootfs INSTALL_MOD_DIR=extras INSTALL_MOD_STRIP=1
+        make -j $(nproc) modules_install SYSSRC=/src INSTALL_MOD_PATH=/rootfs/usr INSTALL_MOD_DIR=extras INSTALL_MOD_STRIP=1
 finalize:
```
dsseng, line 41: Any reason we do not do this test on NVIDIA proprietary?
```suggestion
    test:
      - |
        # https://www.kernel.org/doc/html/v4.15/admin-guide/module-signing.html#signed-modules-and-stripping
        find /rootfs/usr/lib/modules -name '*.ko' -exec grep -FL '~Module signature appended~' {} \+
finalize:
```

kernel/prepare/pkg.yaml
```diff
@@ -11,42 +11,21 @@ steps:
...
-
-        mkdir -p /usr/lib/pkgconfig
-        ln -s /toolchain/include /usr/include
-
         make mrproper
       - |
-        cd /toolchain && git clone https://github.com/a13xp0p0v/kernel-hardening-checker.git
+        mkdir -p /toolchain/kernel-hardening-checker
+        cd /toolchain/kernel-hardening-checker
+        curl -L https://github.com/a13xp0p0v/kernel-hardening-checker/archive/refs/heads/master.tar.gz | tar xz --strip-components=1
```
dsseng, line 24: Downloading this without any tag/hash lock has an advantage of bringing new tests in as soon as possible, however drawbacks are that this is a script being executed in kernel build process right before the kernel build (so if it's compromised it poses a security risk) and older builds (and thus configs) might potentially regress as time passes, due to new checks not accepting old configs

kernel/build/pkg.yaml
```diff
@@ -43,14 +43,6 @@ finalize:
...
-  - from: /toolchain
-    to: /toolchain
-  - from: /usr
-    to: /usr
-  - from: /bin
-    to: /bin
-  - from: /lib
-    to: /lib
+  - from: /
+    to: /
```
dsseng, line 47: Do it this way to avoid losing symlinks. Anyway, these are not released not reproducibility-checked, so it's okay to just pass this stage to the next part as-is

fhs/pkg.yaml
```diff
@@ -1,14 +1,26 @@
...
-    mkdir -pv ${PREFIX}/usr/local/{bin,sbin,lib,include}
-    mkdir -pv ${PREFIX}/var/{log,run,mail,spool}
+    mkdir -pv ${PREFIX}/var/{log,mail,spool}
     install -dv -m 0750 ${PREFIX}/root
     install -dv -m 1777 ${PREFIX}/tmp ${PREFIX}/var/tmp
+    cd ${PREFIX}
+    ln -sT usr/lib lib64
+    ln -sT usr/lib lib
+    ln -sT usr/bin bin
+    ln -sT usr/sbin sbin
```
dsseng, line 20: Merging bin and sbin, as e.g. Arch Linux did, is possible, but is a bit more difficult (as more packages install there) and not necessary to do in this PR since it can build without that

fhs/pkg.yaml
```diff
@@ -1,14 +1,23 @@
...
+    mkdir -pv ${PREFIX}/var/{log,mail,spool}
     install -dv -m 0750 ${PREFIX}/root
     install -dv -m 1777 ${PREFIX}/tmp ${PREFIX}/var/tmp
+    ln -sT lib usr/lib64
+    ln -sT usr/lib lib64
+    ln -sT usr/lib lib
+    ln -sT usr/bin bin
+    ln -sT usr/sbin sbin
+    ln -sT ../run var/run
+    ln -sT ../run/lock var/lock
```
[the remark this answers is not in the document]
dsseng: I verified these packages to match the structure:
`apparmor  containerd  dosfstools  fhs      grub      iptables  kmod    libattr  libinih    liblzma  libnftnl  libseccomp  libsepol  linux-firmware  musl     pcre2  socat          util-linux
cni       cryptsetup  e2fsprogs   flannel  ipmitool  ipxe      libaio  libcap   libjson-c  libmnl   libpopt   libselinux  liburcu   lvm2            openssl  runc   systemd-udevd  xfsprogs`

kernel/prepare/pkg.yaml
smira: right, we probably need a way to separate out `test` so that it doesn't affect steps (I can look into that).

dsseng: I can do that. However what about building older kernel configs in time when there's a breaking change in KSPP checker?

smira: yeah, we can pin it as we discussed in Slack

ipmitool/pkg.yaml
```diff
@@ -16,11 +16,10 @@ steps:
       - |
         tar -xzf ipmitool.tar.gz --strip-components=1
 
+        # ipmitool prefers wget, however only curl supports file:// scheme
```
frezbo, line 19: cool hack to prevent network access :+1:

iptables/pkg.yaml
frezbo: just add a comment on top of patch from where it was obtained from, usually alpine has required patches

kernel/build/pkg.yaml
frezbo: but this would mean we copy a large set of files un-necessarily

frezbo: nevermind we were doing the same earlier

kernel/prepare/pkg.yaml
```diff
@@ -9,44 +9,26 @@ steps:
...
         make mrproper
       - |
-        cd /toolchain && git clone https://github.com/a13xp0p0v/kernel-hardening-checker.git
+        mkdir -p /toolchain/kernel-hardening-checker
+        tar -xzf kspp.tar.gz --strip-components=1 -C /toolchain/kernel-hardening-checker
     install:
       - |
         mkdir -p /src
         cp -a . /src/
+        rm -r /tmp/build/*
```
frezbo, line 31: why are we doing this? We shouldn't rely on bldr internal hacks

lvm2/pkg.yaml
```diff
@@ -41,58 +41,9 @@ steps:
...
-              /sbin/vgimportdevices \
-              /sbin/vgmerge \
-              /sbin/vgmknodes \
-              /sbin/vgreduce \
-              /sbin/vgremove \
-              /sbin/vgrename \
-              /sbin/vgs \
-              /sbin/vgscan \
-              /sbin/vgsplit
+             MODPROBE_CMD=/usr/sbin/modprobe
```
frezbo, line 46: is this where now it ends up on Talos rootfs?

lvm2/pkg.yaml
```diff
@@ -41,58 +41,9 @@ steps:
              --enable-udev_sync \
              --enable-udev_rules \
              --enable-static_link \
+             --prefix=/usr \
              --with-udev-prefix=/usr \
-             MODPROBE_CMD=/sbin/modprobe
-
-        rm -f /sbin/blkdeactivate \
```
frezbo, line 47: not needed anymore?

musl/pkg.yaml
```diff
@@ -43,6 +38,13 @@ steps:
       - |
         cd build
         make DESTDIR=/rootfs install
+
+        mkdir -p /rootfs/usr/bin /rootfs/usr/lib
+        rm -rf /rootfs/lib
```
frezbo, line 43: add a comment on why this is being done, should help future maintainers

sd-boot/pkg.yaml
```diff
@@ -9,8 +9,8 @@ steps:
       - |
         mkdir -p /rootfs
 
-        cp /toolchain/lib/systemd/boot/efi/*.efi /rootfs
-        cp /toolchain/lib/systemd/boot/efi/*.efi.stub /rootfs
+        cp /usr/lib/systemd/boot/efi/*.efi /rootfs
+        cp /usr/lib/systemd/boot/efi/*.efi.stub /rootfs
```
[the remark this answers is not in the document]
frezbo: yeh, let's have a systemd target and these sd-boot/udevd targets just copies right files from there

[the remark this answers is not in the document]
frezbo: this can be another PR

systemd-udevd/pkg.yaml
```diff
@@ -17,13 +17,10 @@ steps:
         sha512: "{{ .systemd_sha512 }}"
     env:
       SOURCE_DATE_EPOCH: {{ .BUILD_ARG_SOURCE_DATE_EPOCH }}
+  - network: default
```
frezbo, line 20: why is this needed?

frezbo (approved):
minor nits

lvm2/pkg.yaml
dsseng: Should be I believe, let me check. /sbin is a symlink to /usr/sbin now

lvm2/pkg.yaml
dsseng: These aren't even under /rootfs. And yes, compiles just fine, no unnecessary stuff in the result

systemd-udevd/pkg.yaml
dsseng: pip3 is used to install ninja etc. will package those into tools later

iptables/pkg.yaml
dsseng: Did it myself just removing those duplicate includes

kernel/prepare/pkg.yaml
dsseng: Hmm, the original copied most directories but not /tmp. Maybe only wanted directories (and not bin lib etc.) should be added to the /rootfs, and then it'll be layered on top of base when used in kernel/build for example?

lvm2/pkg.yaml
dsseng: nah, it's in /usr/bin

PR #1161, opened 2025-02-19
fix: remove patches and other files from copy-only packages

dsseng:
Prevent /pkg/patches from being in the final rootfs

Also add a test for no extra directories in the root and bump tools base image.


reproducibility/pkg.yaml
```diff
@@ -55,6 +55,20 @@ steps:
       [ -L /usr/lib64 ] && [ -d /usr/lib64 ]
       [ -L /sbin ] && [ -d /sbin ]
       [ -L /usr/sbin ] && [ -d /usr/sbin ]
+    - |
+      # Test for extra files/directories
+      # go and pkg come from base
+      ALLOWED_DIRS="usr etc bin sbin lib lib64 dev proc sys opt run var root tmp home go pkg"
```
dsseng, line 61: Err, we perhaps have /pkg in all the copy stages even when they don't have patches, let me see. Should we maybe also remove /pkg from tools somehow?

Pkgfile
```diff
@@ -5,7 +5,7 @@ format: v1alpha2
 vars:
   TOOLCHAIN_MUSL_IMAGE: ghcr.io/siderolabs/toolchain-musl:v0.14.0-alpha.0-6-g818b320
   TOOLS_PREFIX: ghcr.io/siderolabs/
-  TOOLS_REV: v1.10.0-alpha.0-12-g758d61c
+  TOOLS_REV: v1.10.0-alpha.0-15-gd93b780
```
dsseng, line 8: will revert: no actual difference in terms of what we need for building pkgs

reproducibility/pkg.yaml
frezbo: `/pkg` shouldn;'t end up in `/rootfs` unless there was an explicit copy

frezbo (approved):

PR #1163, opened 2025-02-21
fix: reproducibility tests

frezbo:
Move the test script to each stage so we catch any FHS invalidations early on.

### DIFF
--- a/.kres.yaml
+++ b/.kres.yaml
@@ -4,7 +4,6 @@ spec:
   targets:
     # - non-related to the kernel, in alphabetical order
     - apparmor
-    - base
     - ca-certificates
     - cni
     - containerd
--- a/base/fhs-validator.sh
+++ b/base/fhs-validator.sh
@@ -1,0 +1,35 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+ROOTDIR="${1:-/rootfs}"
+
+DENYLIST_PATHS="lib64 lib bin sbin usr/lib64 usr/sbin"
+
+# make sure there's are no symlinks from DENYLIST_PATHS
+for DENYLIST_PATH in ${DENYLIST_PATHS}; do
+    # first check if the the file exists
+    if [[ -e "${ROOTDIR}/${DENYLIST_PATH}" ]]; then
+        if [[ -L "${ROOTDIR}/${DENYLIST_PATH}" ]]; then
+            echo "Found symlink ${ROOTDIR}/${DENYLIST_PATH} which is not allowed"
+            exit 1
+        fi
+
+        if [[ -d "${ROOTDIR}/${DENYLIST_PATH}" ]]; then
+            echo "Found directory ${ROOTDIR}/${DENYLIST_PATH} which is not allowed"
+            exit 1
+        fi
+    fi
+done
+
+# Test for extra files/directories
+ALLOWED_DIRS="usr etc bin sbin lib lib64 dev proc sys opt run var root tmp home"
+
+find "${ROOTDIR}" -mindepth 1 -maxdepth 1 | while read -r DIR; do
+    RELATIVE_DIR=$(basename "${DIR}")
+
+    if ! echo "${ALLOWED_DIRS}" | grep -q "${RELATIVE_DIR}"; then
+        [[  -d "${DIR}" ]] && echo "${DIR} is not an allowed directory" || echo "${DIR} is not an allowed file"
+        exit 1
+    fi
+done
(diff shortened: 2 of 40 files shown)


.kres.yaml
```diff
@@ -4,7 +4,6 @@ spec:
   targets:
     # - non-related to the kernel, in alphabetical order
     - apparmor
-    - base
```
dsseng, line 7: It's used by extensions as the tools base package. However we might probably want to use tools directly there in place of pkgs/base

sd-boot/pkg.yaml
```diff
@@ -8,10 +8,13 @@ steps:
   - install:
     - |
       cd /tmp/systemd
-      mkdir -p /rootfs/
+      mkdir -p /rootfs/usr/lib/systemd/boot/efi
 
-      cp build/src/boot/*.efi /rootfs/
-      cp build/src/boot/*.efi.stub /rootfs/
+      cp build/src/boot/*.efi /rootfs/usr/lib/systemd/boot/efi
+      cp build/src/boot/*.efi.stub /rootfs/usr/lib/systemd/boot/efi
```
dsseng, line 14: Should we? This package is not installed at a rootfs, but used in Talos' Dockerfile to copy from, so it might be okay to put necessary stuff on the root, and do `COPY --from=pkg-sd-boot /*.efi /sd-boot-${TARGETARCH}.efi` instead of `COPY --from=pkg-sd-boot /usr/lib/systemd/boot/efi/*.efi /sd-boot-${TARGETARCH}.efi`

It's okay to make it exempt from fhs if it's not used like that

grub/pkg.yaml
```diff
@@ -64,6 +64,12 @@ steps:
         mkdir -p /rootfs/usr/share/locale
         mkdir -p /rootfs/usr/share/bash-completion/completions
         mv -v /rootfs/etc/bash_completion.d/grub /rootfs/usr/share/bash-completion/completions
+
+        mv /rootfs/usr/sbin/* /rootfs/usr/bin/
```
dsseng, line 68: Maybe there's an option in configure (or a make option) to tell it where to place the binaries. It'd be better, but unsure if one exists

base/fhs-validator.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+ROOTDIR="${1:-/rootfs}"
+
+# Make sure symlinks from FHS did not get overwritten
+for EXPECTED_SYMLINK in bin lib lib64 usr/lib64 sbin usr/sbin; do
+    if [[ -e "${ROOTDIR}/${EXPECTED_SYMLINK}" && ! -L "${ROOTDIR}/${EXPECTED_SYMLINK}" ]]; then
```
dsseng, line 9: I guess this might falsely report okay if package creates its own symlink for example. Perhaps if this script is used without FHS in the first layer it should check that package does not have any of the paths which are symlinks under its root, something like extensions-validator does (but let's make it a denylist here)

.kres.yaml
frezbo: yeh my intention was to use tools directly there

base/fhs-validator.sh
frezbo: good catch

sd-boot/pkg.yaml
frezbo: yeh, makes sense, keep as is and drop the test

.kres.yaml
dsseng: Should be good

base/fhs-validator.sh
```diff
@@ -0,0 +1,35 @@
...
+
+        if [[ -d "${ROOTDIR}/${DENYLIST_PATH}" ]]; then
+            echo "Found directory ${ROOTDIR}/${DENYLIST_PATH} which is not allowed"
+            exit 1
+        fi
+    fi
+done
+
+# Test for extra files/directories
+ALLOWED_DIRS="usr etc bin sbin lib lib64 dev proc sys opt run var root tmp home"
```
dsseng, line 26: Hmm, actually, here we first check if there's something denied, and then deny everything not allowed. Perhaps a better and clearer approach would be to check for allowlist (not including non-usrmerged /bin ...), and then do an allowlist check on the usr itself (must include nothing more than bin, lib, share, etc, maybe sth else). This test shouldn't be ran on FHS itself since it creates /bin and others as symlinks

grub/pkg.yaml
```diff
@@ -45,6 +45,7 @@ steps:
           esac
 
           ../configure --prefix=/usr \
+                    --sbindir=/usr/bin \
```
dsseng, line 48: :+1:

dsseng (approved):

PR #1164, opened 2025-02-21
feat: update Linux to 6.12.16, validate package structure

dsseng:
Latest LTS kernel.

Also rekres and implement a validator for rules like in https://lists.debian.org/debian-devel-announce/2025/02/msg00004.html to ensure clean packages.


smira (approved):

frezbo (approved):
:cool:

PR #1165, opened 2025-02-21
fix: imager deps

frezbo:
Add zlib and export glib deps.

### DIFF
--- a/glib/pkg.yaml
+++ b/glib/pkg.yaml
@@ -15,7 +15,15 @@ steps:
 
         meson setup \
           --prefix=/usr \
-          -Ddefault_library=both \
+          -Dman-pages=disabled \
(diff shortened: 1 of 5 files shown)


glib/pkg.yaml
```diff
@@ -25,6 +33,9 @@ steps:
     install:
       - |
         DESTDIR=/rootfs ninja -C _build install
+
+        # we only need the headers and libs
+        rm -rf /rootfs/usr/{bin,share}
```
dsseng, line 38: No options to disable mans and utilities?

dsseng (approved):

glib/pkg.yaml
frezbo: nope, glib and qemu are huge, even if we're just building part of it

PR #1174, opened 2025-03-03
feat: nfsd support as a module

cm-enfuse:
Build nfsd support as a module package. Attempting to finish this pre-existing PR:  https://github.com/siderolabs/pkgs/pull/713 
 See comment for referenced PR for the extension repo.


cm-enfuse:
Relevant extension: https://github.com/siderolabs/extensions/pull/630

nfsd/nfsd.config
```diff
@@ -0,0 +1,13 @@
+CONFIG_NFSD=m
```
smira, line 1: I think we can simplify this a bit, as this is an in-tree kernel module, and build it as part of the `kernel` build, and then extract it as an extension

I cant take a stab at it

smira (approved):
refactored this to be a kernel module build with base kernel build

PR #1175, opened 2025-03-04
feat(kernel): enable TPM SPI support in kernel config for arm64 

ehbello:
**Description**:

This merge request enables TPM support via the SPI interface on ARM64 by setting `CONFIG_TCG_TIS_SPI=y` in the kernel configuration file. Previously, this option was commented out, preventing SPI-based TPM modules from being supported.

I'm working on supporting secure boot on a RockPi 4c with Talos and, with this change and others pending, it would be possible to use the TPM SLB9670 module on RK3399 platforms.


smira:
@ehbello there's also another option unlocked https://cateee.net/lkddb/web-lkddb/TCG_TIS_SPI_CR50.html with TIS_SPI, do you need that or not?

smira (approved):

PR #1186, opened 2025-03-13
feat(containerd): provide ctr as part of the default containerd Talos build

redhatrises:
This PR enables the Talos containerd package build to provide ctr as part of the Talos releases. ctr is often used for:

- Allows for easy debugging and troubleshooting container issues at the runtime level
- Provides visibility into certain container events as well as container data

Providing ctr as part of the builds in this repo ensures that the ctr version easily lines up with the released containerd version as well as makes it easier for testing, support, and certification with the Talos builds. Alternatively, it might be preferred to have ctr delivered as an extension instead. So, I will open a PR in the extensions repo in case a ctr extension is preferred instead of providing ctr by default as part of the Talos pkg builds.

### DIFF
--- a/containerd/pkg.yaml
+++ b/containerd/pkg.yaml
@@ -28,7 +28,7 @@ steps:
         mkdir -p /rootfs/usr/bin
         make install PREFIX=/usr DESTDIR=/rootfs
 
-        rm /rootfs/usr/bin/{containerd-stress,ctr}
+        rm /rootfs/usr/bin/containerd-stress
     test:
       - |
         fhs-validator /rootfs


smira:
Can you please give an insight in what exactly is missing in Talos API and Kubernetes API which requires you to use `ctr`? How do you plan to use it? Running `ctr` from e.g. a Kubernetes privileged pod requires `containerd` to be working, so not much in troubleshooting?

There is already (at least):

* `talosctl containers`
* `talosctl images`
* `talosctl restart`
* `talosctl cgroups`
* `talosctl logs`
* all Kubernetes pod APIs
* kubelet API

redhatrises:
> Can you please give an insight in what exactly is missing in Talos API and Kubernetes API which requires you to use `ctr`? How do you plan to use it? Running `ctr` from e.g. a Kubernetes privileged pod requires `containerd` to be working, so not much in troubleshooting?
> 
> There is already (at least):
> 
> * `talosctl containers`
> * `talosctl images`
> * `talosctl restart`
> * `talosctl cgroups`
> * `talosctl logs`
> * all Kubernetes pod APIs
> * kubelet API

Had a meeting with @rothgar several weeks back on this topic as we have a joint customer asking for a joint resolution on this. Our software deploys as a privileged daemonset to every node that runs as a local process that interacts at the container runtime level to get events and visibility that both the talos/k8s apis won't have or see.

rothgar:
I was thinking we should add this as an optional extension that only contains the `ctr` binary. We don't want it by default on the system because ideally people can use the talos system API or Kubernetes API to interact with containers. But for products that exec out to `ctr` or do things outside of those paths we need to have an option to include it.

smira:
Ok, makes sense, we should probably have it outlined pretty clearly in the extension. And using `ctr` directly is not supported in general outside of read-only operations.

smira (approved):

smira:
I'll update Talos to exclude `ctr` now from the base system.

PR #1200, opened 2025-04-03
fix: build Amazon ENA driver as module

smira:
This allows it to be replaced with out-of-tree driver https://github.com/siderolabs/pkgs/pull/1197.


dsseng (approved):

smira:
New module: `kernel/drivers/net/ethernet/amazon/ena/ena.ko`

PR #1201, opened 2025-04-07
feat: enable Hisilicon NIC drivers on arm64

shirok1:
These NICs are used on Huawei's ARM-based servers e.g. TaiShan 200 I'm using.

I basically copied corresponding config on Debian's 6.12 kernel and replaced "m"s with "y"s.

Hopefully this get back-ported to Talos 1.9 branch!


shirok1:
> Let's please make them modules - whatever can be made as modules.

Do we need to make a corresponding Talos extension for it? Or just config as modules and good to go.

Also I just found I'm working on my main branch, I'll close this PR and open another one later.

smira:
> > Let's please make them modules - whatever can be made as modules.
> 
> Do we need to make a corresponding Talos extension for it? Or just config as modules and good to go.
> 
> Also I just found I'm working on my main branch, I'll close this PR and open another one later.

All PRs should go to `main`, no, we don't need to make an extension for it, we ship all network drivers in base Talos.

We just need to make more stuff in `arm64` to be modules, but we aren't there yet.

PR #1213, opened 2025-04-24
feat: enable bcache module

mtb-xt:
Hi team,

I would like to enable the bcache module in the talos kernel, to make a **bcache** extension.
The rationale is that Linstor storage can use bcache  for block device caching, and talos already supports DRBD.

I've added the amd64 module, but for some reason, for arm64 it's already enabled. Here -> https://github.com/siderolabs/pkgs/blob/d04243270a4f10f9ecb889883ab42687e5ae6351/kernel/build/config-arm64#L2712 
Does it need to be enabled? Should it also be a module?


frezbo:
> Hi team,
> 
> I would like to enable the bcache module in the talos kernel, to make a **bcache** extension. The rationale is that Linstor storage can use bcache for block device caching, and talos already supports DRBD.
> 
> I've added the amd64 module, but for some reason, for arm64 it's already enabled. Here ->
> 
> https://github.com/siderolabs/pkgs/blob/d04243270a4f10f9ecb889883ab42687e5ae6351/kernel/build/config-arm64#L2712
> 
> Does it need to be enabled? Should it also be a module?

yes, let's also make the arm64 a module

frezbo (approved):

frezbo:
New modules:

AMD64

```text
 diff  /tmp/a/usr/lib/modules/6.12.24-talos/modules.order /tmp/b/usr/lib/modules/6.12.24-talos/modules.order 
12a13
> kernel/lib/crc64.ko
496a498
> kernel/drivers/md/bcache/bcache.ko
```

ARM64

```text
❯ diff  /tmp/a/usr/lib/modules/6.12.24-talos/modules.order /tmp/b/usr/lib/modules/6.12.24-talos/modules.order
134a135
> kernel/drivers/md/bcache/bcache.ko
```

PR #1218, opened 2025-05-01
fix: downgrade libseccomp to 2.5.5

smira:
See https://github.com/siderolabs/talos/issues/10855

containerd-2.0 is using 2.5.4: https://github.com/containerd/containerd/blob/release/2.0/script/setup/install-seccomp

main (2.1) is using 2.5.5: https://github.com/containerd/containerd/blob/main/script/setup/install-seccomp

2.6.0 might be too new/unsuported.


smira:
I have a test running right now with 2.5.5 to confirm (or not) if that fixes crashes.

frezbo (approved):

PR #1219, opened 2025-05-02
fix: containerd build doesn't need seccomp

smira:
libseccomp is only used for runc build.


dsseng (approved):

smira:
Ref: https://github.com/containerd/containerd/blob/release/2.0/BUILDING.md#build-requirements

frezbo (approved):

PR #1220, opened 2025-05-02
fix: build containerd with Go 1.23

smira:
Fixes https://github.com/siderolabs/talos/issues/10855


smira:
Note: for Talos main, we can switch to Go 1.24.3 which has a fix, for Talos 1.10 we can have this fix till the EOL.

frezbo (approved):

PR #1232, opened 2025-05-15
feat: Enable CONFIG_EROFS_FS in kernel config

jnohlgard:
This is potentially useful for IMAGECACHE partitions and as an alternative for squashfs for the rootfs.

> ## EROFS - Enhanced Read-Only File System
> 
> A modern, flexible, general-purpose, high-performance block-based immutable filesystem with highly-optimized ondisk format and runtime implementation aimed at various use cases instead of just focusing on storage space saving without considering any side effects of runtime performance.
> 

See: https://erofs.docs.kernel.org/en/latest/


smira:
We looked into that, but there's no usecase right now for it in Talos core. Anything you wanted to use it for?

smira:
Talos doesn't support EROFS in any way for the image cache today, and the plan is to allow image cache to be updated on the fly (e.g. by inserting new image cache volume).

PR #1235, opened 2025-05-19
fix: bring back updated containerd gvisor patch

smira:
When updating for containerd 2.1, I assumed the patch is no longer needed, as it was included into containerd, but it turns out it got reverted upstream: https://github.com/containerd/containerd/pull/11793

So bring back the patch in updated form, as otherwise `gvisor` leaves running processes behind on container stop.

### DIFF
--- a/containerd/patches/11741-not-set-sandbox-id-when-use-podsandbox-type.patch
+++ b/containerd/patches/11741-not-set-sandbox-id-when-use-podsandbox-type.patch
@@ -1,0 +1,65 @@
+From e9804ee0e9d85788648b589c17e67a024a93151e Mon Sep 17 00:00:00 2001
+From: yylt <yang8518296@163.com>
+Date: Tue, 22 Apr 2025 11:05:54 +0800
+Subject: [PATCH] not set sandbox id when use podsandbox type
+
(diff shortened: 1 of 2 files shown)


frezbo (approved):

samuelkarp:
https://github.com/containerd/containerd/pull/11793 is a new fix for the same issue, so the previous patch should no longer be needed.  If you are still running into problems with gVisor and containerd, please open an issue with the containerd project.

smira:
> [containerd/containerd#11793](https://github.com/containerd/containerd/pull/11793) is a new fix for the same issue, so the previous patch should no longer be needed. If you are still running into problems with gVisor and containerd, please open an issue with the containerd project.

thanks, will do this after I confirm that it indeed fixes the issue. I could only observe that some process is left over in the host after gvisor-backed is terminated (previously it was stuck on termination), but now the pod is terminated, but something is left running (need to check more).

smira:
> [containerd/containerd#11793](https://github.com/containerd/containerd/pull/11793) is a new fix for the same issue, so the previous patch should no longer be needed. If you are still running into problems with gVisor and containerd, please open an issue with the containerd project.

@samuelkarp created an issue https://github.com/containerd/containerd/issues/11871

PR #1246, opened 2025-05-23
chore: remove unused udev programs

BobyMCbobs:
removes programs in /usr/lib/udev which are not called from udev rules.

Steps to reproduce findings
```
git clone https://github.com/siderolabs/talos 
cd talos
make initramfs
cd _out
zstd -d ./initramfs-amd64.xz
cpio -iv < initramfs-amd64
mkdir ./squashy
sudo mount --type squashfs --options loop --source ./rootfs.sqsh --target ./squashy
cd ./squashy/usr/lib/udev/rules.d
find .. -maxdepth 1 -mindepth 1 -type f -executable | sed 's,../,,g' | xargs -n1 sh -c 'grep -q -Rnw $PWD -e $1 || echo $1' sh
```

### DIFF
--- a/systemd-udevd/pkg.yaml
+++ b/systemd-udevd/pkg.yaml
@@ -31,6 +31,7 @@ steps:
         # Already built this
         rm -rf /rootfs/usr/lib/udev/hwdb.d
         rm /rootfs/usr/lib/udev/rules.d/{README,60-cdrom_id.rules,60-persistent-alsa.rules,60-persistent-v4l.rules,64-btrfs.rules,70-joystick.rules,70-mouse.rules,70-touchpad.rules,78-sound-card.rules,90-vconsole.rules,99-systemd.rules}
+        rm -f /rootfs/usr/lib/udev/{cdrom_id,v4l_id}
         # Azure csi support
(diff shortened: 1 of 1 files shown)


systemd-udevd/pkg.yaml
```diff
@@ -31,6 +31,7 @@ steps:
         # Already built this
         rm -rf /rootfs/usr/lib/udev/hwdb.d
         rm /rootfs/usr/lib/udev/rules.d/{README,60-cdrom_id.rules,60-persistent-alsa.rules,60-persistent-v4l.rules,64-btrfs.rules,70-joystick.rules,70-mouse.rules,70-touchpad.rules,78-sound-card.rules,90-vconsole.rules,99-systemd.rules}
+        rm -f /rootfs/usr/lib/udev/{cdrom_id,v4l_id}
```
frezbo, line 34: @dsseng think these are needed anywhere?

dsseng: Hmm, IIRC those are called by some of the built-in rules on some occasions

smira: yeah, I think they are being used

PR #1262, opened 2025-06-10
feat: add hailort package

michael-robbins:
Add kernel module for hailort Hailo-8L (and others in the family) hardware.


smira (approved):
thank you!

PR #1265, opened 2025-06-18
feat: add SBOM metadata for some packages

dsseng:

### DIFF
--- a/apparmor/pkg.yaml
+++ b/apparmor/pkg.yaml
@@ -36,6 +36,14 @@ steps:
     test:
       - |
         fhs-validator /rootfs
+    sbom:
+      outputPath: /rootfs/usr/share/spdx/apparmor.spdx.json
+      version: {{ .apparmor_version }}
+      cpes:
+        - cpe:2.3:a:apparmor:apparmor:{{ .apparmor_version }}:*:*:*:*:*:*:*
+        - cpe:2.3:a:canonical:apparmor:{{ .apparmor_version }}:*:*:*:*:*:*:*
+      licenses:
+        - GPL-2.0-or-later
 finalize:
   - from: /rootfs
     to: /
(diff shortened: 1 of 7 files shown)


apparmor/pkg.yaml
```diff
@@ -36,6 +36,14 @@ steps:
     test:
       - |
         fhs-validator /rootfs
+    sbom:
+      outputPath: /rootfs/usr/share/spdx/apparmor.spdx.json
+      version: {{ .apparmor_version }}
+      cpes:
+        - cpe:2.3:a:apparmor:apparmor:{{ .apparmor_version }}:*:*:*:*:*:*:*
+        - cpe:2.3:a:canonical:apparmor:{{ .apparmor_version }}:*:*:*:*:*:*:*
```
dsseng, line 44: Not all packages have reliable well-known PURLs I could find

nonfree/kmod-nvidia/production/pkg.yaml
```diff
@@ -43,6 +43,15 @@ steps:
...
         fhs-validator /rootfs
+    sbom:
+      outputPath: /rootfs/usr/share/spdx/kmod-nvidia-production.spdx.json
+      version: {{ .nvidia_driver_production_version }}
+      cpes:
+        - cpe:2.3:a:nvidia:gpu_display_driver:{{ .nvidia_driver_production_version }}:*:*:*:*:*:*:*
+        - cpe:2.3:a:nvidia:gpu_display_driver:{{ .nvidia_driver_production_version }}:*:*:*:*:linux_kernel:*:*
+      purl: pkg:nvidia/gpu-display-driver@{{ .nvidia_driver_production_version }}
+      licenses:
+        - CUSTOM
```
dsseng, line 54: Not sure how should we put that. Either custom, EULA or maybe specifically `CUSTOM-NVIDIA`. But that should certainly signal that this is not a universally known license but a vendor-specific one.

frezbo: does any vendor ship these info?

dsseng: https://archlinux.org/packages/extra/x86_64/nvidia/ says it's LicenseRef-custom

smira (approved):
let's try with these SPDX, and close the loop on Talos side to build the combined one, and put it to the rootfs/release assets - once it's done, we can go back and add more SBOMs

PR #1304, opened 2025-08-21
chore: update kernel config to support 8192 CPUs

SoulKyu:
Update kernel configuration to increase maximum CPU support from 512 to 8192. This allows Talos to run on modern high-core-count systems.

Changes:
- Update CONFIG_NR_CPUS_RANGE_END from 512 to 8192
- Update CONFIG_NR_CPUS from 512 to 8192

This change supports modern server hardware with >512 CPUs while maintaining backward compatibility, enabling support for significantly more CPUs.

Fixes support for systems with more than 512 CPUs.

Here is our actual node description : 
```
Capacity:
  cpu:                512
  ephemeral-storage:  193225588Ki
  hugepages-1Gi:      0
  hugepages-2Mi:      0
  memory:             1188355220Ki
Allocatable:
  cpu:                511950m
  ephemeral-storage:  177808266150
  hugepages-1Gi:      0
  hugepages-2Mi:      0
  memory:             1187859604Ki
  ```
  
  Our host has more than 512 CPU.
  
  From the Kernel documentation : 
  ```
 This allows you to specify the maximum number of CPUs which this
kernel will support. If CPUMASK_OFFSTACK is enabled, the maximum
supported value is 8192, otherwise the maximum value is 512. The
minimum value which makes sense is 2.

This is purely to save memory: each supported CPU adds about 8KB
to the kernel image. 
```

Which lead to approximately +65Mb to the image size.


smira (approved):
thank you

PR #1315, opened 2025-08-29
Enable USB audio support

jonkerj:
I want to run [librespot](https://github.com/librespot-org/librespot) on my home Talos cluster using an USB audio device; this commit adds the kernel support for that.


smira:
@jonkerj just clarifying - is your plan to introduce a system extension with these drivers?

jonkerj:
> @jonkerj just clarifying - is your plan to introduce a system extension with these drivers?

Yeah

smira (approved):

PR #1338, opened 2025-09-29
feat: upgrade Linux to 6.17

dsseng:
Newest stable series


kernel/build/config-amd64
```diff
@@ -1631,7 +1637,6 @@ CONFIG_IP_VS_NFCT=y
 # IP: Netfilter Configuration
 #
 CONFIG_NF_DEFRAG_IPV4=y
-CONFIG_IP_NF_IPTABLES_LEGACY=y
```
dsseng, line 1634: Dropping legacy iptables, hopefully nothing is using it due to conflict with nftables which is in use by Talos

frezbo:
> > any kernel module change in paths we need to track?
> 
> If they only need to be considered for Talos side, then I could merge and do the diff on the CI-built artifacts to save build time

yeh, that's fine, I do that mostly, then add a comment here once it's built

kernel/build/config-amd64
```diff
@@ -7343,6 +7343,7 @@ CONFIG_TRACING=y
 CONFIG_GENERIC_TRACER=y
 CONFIG_TRACING_SUPPORT=y
 CONFIG_FTRACE=y
+CONFIG_TRACEFS_AUTOMOUNT_DEPRECATED=y
```
frezbo, line 7346: drop this and update the mounts in talos

dsseng: we already mount `/sys/kernel/tracing`, so perhaps no change is needed

PR #1356, opened 2025-10-22
Enable CONFIG_PCI_P2PDMA for GPUDirect RDMA

nicolerenee:
This enables `CONFIG_PCI_P2PDMA` to support GPUDirect RDMA using DMA-BUF. 

NVIDIA recommends using DMA-BUF instead of nvidia-peermem for GPUDirect RDMA. This enables peer-to-peer DMA between PCI devices, allowing:

- GPU memory directly accessible to network adapters
- GPU-to-GPU transfers without CPU involvement
- GPU-to-storage direct paths

Originally requested in siderolabs/talos#12021


kernel/build/config-amd64
```diff
@@ -1983,7 +1983,7 @@ CONFIG_PCI_NPEM=y
 CONFIG_PCI_PRI=y
 CONFIG_PCI_PASID=y
 # CONFIG_PCIE_TPH is not set
-# CONFIG_PCI_P2PDMA is not set
+CONFIG_PCI_P2PDMA=y
```
frezbo: can we make this a module?

smira: yes, I think so.

PR #1397, opened 2025-12-09
feat: add mstflint module

jfroy:


smira:
> > ERROR: failed to build: failed to solve: digest mismatch sha256:3590bcb90a75c32ba8b10d692d26838caedbc267a57db23931694abc9598c873: sha256:b70576ee89adb5f0be404b0d4e714623b10d6d9d1a041ebdf11f8a423f95be1d
> 
> I don't think this is caused by the PR?

yes, this is GitHub having issue.

PR #1411, opened 2025-12-17
feat: enable CONFIG_MISC_RP1 in ARM64 config

eseiker:
Raspberry Pi 5 requires MISC_RP1 drivers to work with PCI peripherals properly


smira (approved):
👍

PR #1423, opened 2025-12-23
feat: update systemd to 259

dsseng:
systemd 259 now officially supports musl libc - select it,
drop patches except for the one removing unneded doc files.


smira (approved):

frezbo (approved):
🆒 🚀

PR #1426, opened 2025-12-27
feat: add realtek 8127 driver on arm

rothgar:
Needed for NVIDIA DGX Spark on-board nic


smira:
This kconfig option doesn't exist, and R8169 handles 8127 in the latest kernel.

rothgar:
Based on what I found from other distros the 8269 driver is actually the problem with the network not coming back up after a reboot.

https://forums.developer.nvidia.com/t/has-anyone-tried-an-alternative-linux-distro/349124/21

I'll have to see if I can find their configuration on why they have that module option.

smira:
There's no such option as R8127 at all, so this PR is not correct anyways.

There's an out of tree 8127 driver, but it's not clear whether it's actually any useful, as R8169 handles now 8127 as well.

PR #1433, opened 2026-01-06
Add lvmd 

pranav767:
Add lvmd as a talos package, allowing CSI drivers to manage logical volumes without direct access to host block devices.

CSI components communicate with lvmd over a gRPC interface and lvmd performs actual LVM operations using the host's lvm2 tools. This avoids the need to run privileged CSI containers or mount host devices like `/dev` for supporting LVM (Creating LV's using CSI)

Reference : https://github.com/topolvm/topolvm/blob/main/docs/lvmd.md
This PR helps to continue work in : https://github.com/siderolabs/talos/pull/12530

### DIFF
--- a/.kres.yaml
+++ b/.kres.yaml
@@ -40,6 +40,7 @@ spec:
     - liburcu
     - linux-firmware
     - lvm2
+    - lvmd
     - mtools
     - musl
     - nftables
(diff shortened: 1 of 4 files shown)


frezbo:
> Yes, had no idea about [extensions](https://github.com/siderolabs/extensions), Yes I think adding lvmd here won't start the service, It needs to be in extensions. I'm still thinking about the design and use case scenarios for LVM, Closing this PR (can create PR in `extensions`, after deciding on a solid design.

Start a github disussion first with ideas/implementation details etc,then once it's good can be converted to an issue and then worked on, avoids wasting time for everyone

PR #1443, opened 2026-01-17
feat: enable kernel irq time accounting

neutralalice:
I am interested in collecting interrupt request metrics within talos. I believe this is the only kernel config definition needed to do so. This definition usually generates a /proc/pressure/irq file which can be ingested in to a tsdb.

Considerations: There is a small performance hit according to sources online, I am unsure if talos has already trialed this config option and evaluated the hit. Some other distributions do build with this option (RHEL notably).


frezbo (approved):
it would be nice to see some benchmarks done by someone, but i think it's okay for this to be merged

PR #1464, opened 2026-02-17
feat: add patch for Cilium BPF verifier rejection by the kernel

nberlee:
This fixes https://github.com/siderolabs/talos/issues/12726


dsseng:
The patch has been submitted just today and hasn't received any reviews on the LKML yet. It is a valuable fix, and should be considered later when there's some feedback on LKML (there should be some by the time this needs to be considered, i.e. before 1.12.5 release and backports for it)

frezbo (approved):
🆒

dsseng:
> @dsseng it looks like the commit you pushed was not signed with your gpg key, also it was not signedoff by you (not sure if the last one is require for siderolabs team members)

1. Not merging yet, I just wanted to run CI. Whoever begins the merge will sign it
2. No need afaik, I've not written or modified any code in the commit, I just rebased your commit on top of the recent kernel-related changes to make sure it's tested with the LLVM toolchain

nberlee:
@frezbo the pull request  referenced does not seem to update the kernel?

PR #1485, opened 2026-03-14
feat: bump deps

frezbo:
Bump dependencies

drbd           - 9.3.0 -> 9.3.1
e2fsprogs      - v1.47.3 -> v1.47.4
systemd        - 259.3 -> 259.5
ipxe           - 9413865 -> 88d3967
libarchive     - 3.8.5 -> 3.8.6
px-fuse        - 95261b1 -> 4b7ade7
glib           - 2.87.3 -> 2.87.5
socat          - 1.8.0.3 -> 1.8.1.1
gdrcopy        - v2.5.1 -> v2.5.2


smira (approved):
Libseccomp should not be bumped (see comment), and probably we should split it up into backportable part?

PR #1508, opened 2026-04-13
kernel: enable CONFIG_UHID and CONFIG_INPUT_JOYDEV as modules

ansgardahlen:
## Summary

Enable `CONFIG_UHID=m` and `CONFIG_INPUT_JOYDEV=m` in the amd64 kernel configuration so that the corresponding system extensions can be built and shipped.

## Use case

Game streaming on Talos Linux using [Sunshine](https://github.com/LizardByte/Sunshine) (server) and [Moonlight](https://github.com/moonlight-stream) (client) requires two kernel modules that are currently disabled:

- **uhid** (`CONFIG_UHID`): Provides `/dev/uhid`, which allows user-space applications to create virtual HID devices. Sunshine uses this to emulate gamepads from Moonlight client input.

- **joydev** (`CONFIG_INPUT_JOYDEV`): Provides `/dev/input/js*` joystick interface. Required by the browser Gamepad API (Chromium) to detect virtual controllers.

Both modules are small, self-contained, and have no additional dependencies.

## What was changed

In `kernel/build/config-amd64`:

```diff
-# CONFIG_UHID is not set
+CONFIG_UHID=m

-# CONFIG_INPUT_JOYDEV is not set
+CONFIG_INPUT_JOYDEV=m
```

## Testing

- Built both modules from kernel source (6.18.18) against the Talos kernel config on a v1.12.6 cluster.
- Packaged as a custom system extension, installed via `imager`.
- Modules loaded successfully but were rejected by `module.sig_enforce=1` (expected — only Siderolabs-signed modules are accepted).
- With enforcement disabled: verified `/dev/uhid` and `/dev/input/js*` creation and Sunshine gamepad passthrough end-to-end.

## Related

Companion PR to `siderolabs/extensions` adding `uhid` and `joydev` system extension definitions (same pattern as existing `uinput` extension).


frezbo:
@ansgardahlen could you signoff the commit via `git commit -s --amend --no-edit && git push --force-with-lease`

frezbo (approved):

frezbo:
I enabled the same for arm64 too

PR #1511, opened 2026-04-14
feat: update containerd to 2.2.3

smira:
See https://github.com/containerd/containerd/releases/tag/v2.2.3


frezbo (approved):
don;t we need for 1.12?

smira:
> don;t we need for 1.12?

1.12 needs it, but it needs 2.1.x bump: https://github.com/siderolabs/pkgs/blob/a92bed5bdd0a3f14312218dd63d6822e38e6dedd/Pkgfile#L16

PR #1166, opened 2025-02-23
(WIP)feat: add l4t package

mmalyska:
Add support for Jetson Orin SBC. Those are modules and display drivers for tegra chips.

## Problem:
I don't see module host1x  to be loaded from the extension.
image
The module host1x is already build in inside kernel 
image
so it cannot be replaced by extension module with the same name host1x.

#### Create custom builder buildx
```bash
docker buildx create --driver docker-container  --driver-opt network=host --name local1 --buildkitd-flags '--allow-insecure-entitlement security.insecure' --use
```
#### Run local doker registry + ui
```
crane export 127.0.0.1:5005/siderolabs/nvidia-l4t:36.4.3-1.9.5-beta extension-nvidia-l4t
cat profile.yaml | docker run --network host --rm -i \
-v $PWD/_out:/out -v $PWD/extension-nvidia-l4t:/extension-nvidia-l4t \
127.0.0.1:5005/siderolabs/imager:1.9.5-beta -
docker login ghcr.io -u mmalyska --password-stdin
docker load < _out/installer-arm64.tar
docker tag 127.0.0.1:5005/siderolabs/installer:1.9.5-beta ghcr.io/mmalyska/talos-images:v1.9.5-beta2
docker push ghcr.io/mmalyska/talos-images:v1.9.5-beta2
```
#### Building the pkgs
```
cd siderolabs-pkgs
make base PLATFORM=linux/arm64,linux/amd64 REGISTRY=127.0.0.1:5005 PUSH=true CI_ARGS="--cache-to=mode=max,type=registry,ref=127.0.0.1:5005/siderolabs/base:cache --cache-from=type=registry,ref=127.0.0.1:5005/siderolabs/base:cache" IMAGE_TAG=1.9.5-beta
make kernel PLATFORM=linux/arm64,linux/amd64 REGISTRY=127.0.0.1:5005 PUSH=true CI_ARGS="--cache-to=mode=max,type=registry,ref=127.0.0.1:5005/siderolabs/kernel:cache --cache-from=type=registry,ref=127.0.0.1:5005/siderolabs/kernel:cache" IMAGE_TAG=1.9.5-beta
[...]

PR #1518, opened 2026-04-22
feat: add nvidia-tegra-nvgpu package for Jetson Orin (GA10B)

schwankner:
## Summary

Adds the OE4T-patched GPU driver stack for NVIDIA Jetson Orin NX (Tegra234 / GA10B) as a Talos kernel package.

**Hardware:** Jetson Orin NX 16GB (Tegra234 / GA10B)
**Tested with:** Talos v1.13, kernel 6.18, CUDA 12.6

### Modules

| Module | Source | Purpose |
|--------|--------|---------|
| `host1x.ko` | OE4T linux-nv-oot (wip-r36.5-take-2) | Syncpoint API with GA10B support + syncpt[0] reservation fix |
| `host1x-fence.ko` | OE4T linux-nv-oot | DMA fence bridge for syncpoints |
| `nvmap.ko` | OE4T linux-nv-oot | GPU memory allocator |
| `mc-utils.ko` | OE4T linux-nv-oot | EMC frequency management |
| `nvhost-ctrl-shim.ko` | embedded `.c` (this pkg) | `/dev/nvhost-ctrl` for JetPack 6 CUDA runtime |
| `nvgpu.ko` | OE4T linux-nvgpu | Main GA10B GPU driver |

### nvhost-ctrl-shim

The JetPack 6 CUDA runtime (`libnvrm_host1x.so`) requires `/dev/nvhost-ctrl` for `cudaStreamSynchronize`. Without it, CUDA falls back to CPU semaphore polling (~7 tok/s vs ~60 tok/s for LLM inference on Orin NX 16GB).

The shim bridges 8 ioctls (`GET_VERSION`, `SYNCPT_READ`, `SYNCPT_READ_MAX`, `SYNCPT_WAITMEX`, `SYNC_FENCE_CREATE`, `GET_CHARACTERISTICS`, `POLL_FD_CREATE`, `SYNC_FILE_EXTRACT`) to the OE4T host1x syncpoint API using `dma_fence` for interrupt-driven waits.

The source (`nvhost_ctrl_shim.c`) is embedded directly in the package directory and made available at `/pkg/` by bldr at build time.

### host1x shadow path

OE4T `host1x.ko` is installed at `kernel/drivers/gpu/host1x/` to shadow the in-tree module. The squashfs overlay replaces the built-in host1x so nvgpu's `CONFIG_TEGRA_GK20A_NVHOST=y` path gets the OE4T version with `HOST1X_SYNCPT_GPU` support and the GA10B ERRATA_SYNCPT_INVALID_ID_0 fix.

### Build notes

- Built with Clang (`LLVM=1`), matching the Talos toolchain
- OE4T linux-nv-oot branch `wip-r36.5-take-2` adds kernel 6.18 compatibility fixes (`__assign_str`, `f_ref`, `__alloc_pages_bulk` 5-arg)
- Module signing uses the kernel-build stage's auto-generated `certs/signing_key.pem`
- `CONFIG_GCC_PLUGIN_LATENT_ENTROPY` stripped from `auto.conf` / `autoconf.h` before OOT builds (Clang compat)
- CI-validated with ARM64 build (~90 min cold / ~20 min warm kernel cache), all 6 `.ko` files confirmed present

### Result

Full CUDA inference in Kubernetes pods on Talos Linux without `privileged: true`, using CDI device injection. Verified ~60 tok/s on `qwen2.5:0.5b` (Ollama) on Jetson Orin NX 16GB.

## References

- Continues: #1166 (WIP l4t pkg — established that modules need to be a pkg first)
- Discussion: siderolabs/sbc-jetson#23 (frezbo/smira confirmed shim belongs in pkgs)
- Extension PR (follow-up, once this merges): siderolabs/extensions#624


nvidia-tegra-nvgpu/pkg.yaml
```diff
@@ -0,0 +1,390 @@
...
+        make -j $(nproc) ARCH=arm64 \
+          src=/oot-src/out/nvidia-conftest/nvidia \
+          obj=/oot-src/out/nvidia-conftest/nvidia \
+          LLVM=1 \
+          NV_KERNEL_SOURCES=/src \
+          NV_KERNEL_OUTPUT=/src \
+          -f /oot-src/out/nvidia-conftest/nvidia/Makefile
+        echo "conftest done."
+      - |
+        # Patch OOT module Makefiles: remove -Werror and add required include paths.
```
frezbo: i think we can move this to a fixup.sh under scripts directory and invoke /pkg/scripts/fixup.sh

schwankner: Done — extracted to `scripts/fixup.sh` and invoked as `/pkg/scripts/fixup.sh` in the build step. The script follows the same pattern as `nvhost_ctrl_shim.c` (package-dir files available at `/pkg/` during bldr build).

nvidia-tegra-nvgpu/pkg.yaml
```diff
@@ -0,0 +1,267 @@
...
+        echo "=== Building governor-pod-scaling ==="
+        make -j $(nproc) ARCH=arm64 \
+          -C /src \
+          M=${NVIDIA_OOT}/drivers/devfreq \
+          CC=/usr/local/bin/clang-oot \
+          CONFIG_TEGRA_OOT_MODULE=m \
+          srctree.nvidia-oot=${NVIDIA_OOT} \
+          srctree.nvconftest=${NVIDIA_CONFTEST} \
+          LLVM=1 \
+          modules 2>&1 && echo "governor-pod-scaling: OK" || echo "governor-pod-scaling: FAILED (continuing)"
```
frezbo: shouldn't we fail here?

nvidia-tegra-nvgpu/pkg.yaml
```diff
@@ -0,0 +1,267 @@
...
...UFRfSU5WQUxJRF9JRF8wIHJlamVjdHMgaWQ9MDsgZHVyaW5nIH4xLTJtcyBhZnRlciBmaXJzdCBrZXJuZWwgc3VibWl0CgkgKiBob3N0MXhfc3luY3B0X2FsbG9jIG1heSByZXR1cm4gTlVMTCwgY2F1c2luZyBDVURBIGVycm9yIDk5OSBvbiBjdWRhU3RyZWFtU3luY2hyb25pemUuICovCglmb3IgKHJldHJ5ID0gMDsgcmV0cnkgPCA1OyByZXRyeSsrKSB7CgkJaG9zdDF4ID0gcGxhdGZvcm1fZ2V0X2RydmRhdGEobnZob3N0X2Rldi0+aG9zdDF4X3BkZXYpOwoJCWlmICghaG9zdDF4KSB7CgkJCXByX3dhcm5fcmF0ZWxpbWl0ZWQoIm52Z3B1OiBob3N0MXggbm90IHJlYWR5LCBzeW5jcHQgcmV0cnkgJWQvNVxuIiwgcmV0cnkgKyAxKTsKCQkJbXNsZWVwKDEpOwoJCQljb250aW51ZTsKCQl9CgkJc3AgPSBob3N0MXhfc3luY3B0X2FsbG9jKGhvc3QxeCwKCQkJCUhPU1QxWF9TWU5DUFRfQ0xJRU5UX01BTkFHRUQgfCBIT1NUMVhfU1lOQ1BUX0dQVSwKCQkJCXN5bmNwdF9uYW1lKTsKCQlpZiAoc3ApCgkJCWJyZWFrOwoJCXByX3dhcm5fcmF0ZWxpbWl0ZWQoIm52Z3B1OiBzeW5jcHRfYWxsb2MgTlVMTCwgcmV0cnkgJWQvNVxuIiwgcmV0cnkgKyAxKTsKCQltc2xlZXAoMSk7Cgl9CglpZiAoIXNwKSB7CgkJcHJfZXJyX3JhdGVsaW1pdGVkKCJudmdwdTogZ2V0X3N5bmNwdF9jbGllbnRfbWFuYWdlZDogZmFpbGVkIGFmdGVyIHJldHJpZXNcbiIpOwoJCXJldHVybiAwOwoJfQoJcmV0dXJuIGhvc3QxeF9zeW5jcHRfaWQoc3ApOwp9').decode();assert OLD in c,'nvhost_host1x.c: pattern not found';c=c.replace(OLD,NEW,1);open(fname,'w').write(c);print('Patched nvhost_host1x.c: retry loop in nvgpu_nvhost_get_syncpt_client_managed')"
```
frezbo: move this as a python script, makes reviewing easier

nvidia-tegra-nvgpu/pkg.yaml
```diff
@@ -0,0 +1,267 @@
...
+
+        # host1x-fence + nvhost-ctrl-shim + all other nvidia-oot modules go to extra/
+        find /oot-src/nvidia-oot/drivers/gpu/host1x-fence -name "*.ko" \
+          -exec install -m 644 {} /rootfs/usr/lib/modules/${KERNEL_RELEASE}/extra/nvidia-tegra/ \;
+        find /oot-src/nvhost-ctrl-shim -name "*.ko" \
+          -exec install -m 644 {} /rootfs/usr/lib/modules/${KERNEL_RELEASE}/extra/nvidia-tegra/ \;
+        find /oot-src/nvidia-oot -name "*.ko" \
+          -exec install -m 644 {} /rootfs/usr/lib/modules/${KERNEL_RELEASE}/extra/nvidia-tegra/ \;
+
+        # Sign all modules with the kernel's signing key (auto-generated by kernel-build stage)
```
frezbo: I guess the OOT doesn't have native signing, usually `make install INSTALL_MOD_STRIP=1 CONFIG_MODULE_SIG_ALL=y` should work

Pkgfile
```diff
@@ -288,5 +288,16 @@ vars:
   gdrcopy_version: v2.5.2
   gdrcopy_sha256: 32bc7b2c198dd97ec251de0ff4823252c95e31a4c79a5f843c82514c9af2052b
   gdrcopy_sha512: c717f118eff8cd5a8dc35613c3881818f8b71dc493461dd0151ce7c882f8e2c2d852e22733fab4e2bec57219e10eec874c11b4fad90dd4815ae572840ed19d28
+
+  # OE4T (NVIDIA Tegra) kernel modules for Jetson Orin NX (Tegra234 / GA10B)
```
frezbo, line 314: could you also add the renovate tags, or at least add a comment where these commits come from, so we get updates

frezbo:
@claude review

nvidia-tegra-nvgpu/pkg.yaml
frezbo: similarly other places where we do `|| true`

schwankner: Fixed — all OOT module build commands now use `|| { echo "BUILD FAILED"; exit 1; }`. If any module (including governor_pod_scaling) fails to compile, the build exits immediately.

The only remaining `|| true` is on the `Module.symvers` accumulation lines (`cat ... >> Module.symvers.nvidia 2>/dev/null || true`) — kept intentionally since a missing symvers entry is non-fatal (cross-module symbol resolution is best-effort).

nvidia-tegra-nvgpu/pkg.yaml
schwankner: Done — extracted to `scripts/patch_nvhost_host1x.py`. The script searches for the exact function signature and replaces it with the retry loop, exiting with a clear error if the pattern is not found.

nvidia-tegra-nvgpu/pkg.yaml
schwankner: Switched to `modules_install INSTALL_MOD_PATH=/rootfs/usr INSTALL_MOD_STRIP=1 CONFIG_MODULE_SIG_ALL=y` for each module directory — the standard KBUILD mechanism handles strip-then-sign in the correct order. host1x uses `INSTALL_MOD_DIR=kernel/drivers/gpu/host1x` (shadow path for in-tree replacement), all others use `INSTALL_MOD_DIR=extra/nvidia-tegra`. Added a `test:` section with `grep -FL '~Module signature appended~'` and `fhs-validator`, matching the pattern in gasket-driver/zfs/hailort. Also fixed `finalize: to: /` (was incorrectly `/rootfs`).

Pkgfile
schwankner: Added `# renovate: datasource=git-refs versioning=git depName=<url>` above each of the three OE4T commit variables. sha256/sha512 will need to be updated manually when commits are bumped (the tarball hash changes with each new commit).

nvidia-tegra-nvgpu/pkg.yaml
```diff
@@ -0,0 +1,267 @@
...
+        echo "Signing kernel modules..."
+        find /rootfs/usr/lib/modules/${KERNEL_RELEASE}/ \
+             -name "*.ko" 2>/dev/null | while read ko; do
+          /src/scripts/sign-file sha512 \
+            /src/certs/signing_key.pem \
+            /src/certs/signing_key.x509 \
+            "${ko}"
+          echo "  Signed: ${ko}"
+        done
```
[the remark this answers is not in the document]
schwankner: Not an issue in the pkgs context: no other kernel module package in this repo (gasket-driver, zfs, hailort, xdma-driver, etc.) runs `depmod` from the build. The `depmod` database is rebuilt by the Talos extension system at boot time after all module overlays are applied. Running it here against a cross-compiled rootfs would produce incorrect results anyway.

nvidia-tegra-nvgpu/pkg.yaml
```diff
@@ -0,0 +1,267 @@
...
+          srctree.nvconftest=/oot-src/out/nvidia-conftest \
+          LLVM=1 \
+          KBUILD_EXTRA_SYMBOLS=/oot-src/out/nvidia-linux-header/Module.symvers.nvidia \
+          KBUILD_MODPOST_WARN=1 \
+          modules
+        echo "nvgpu done."
+    install:
+      - |
+        KERNEL_RELEASE=$(cat /src/include/config/kernel.release)
+        echo "Installing for kernel ${KERNEL_RELEASE}..."
```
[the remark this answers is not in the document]
schwankner: Fixed — a `test:` section has been added with both checks:
```
find /rootfs/usr/lib/modules -name '*.ko' -exec grep -FL '~Module signature appended~' {} \+
fhs-validator /rootfs
```
Matches the pattern in gasket-driver, zfs, hailort, and others.

nvidia-tegra-nvgpu/pkg.yaml
```diff
@@ -0,0 +1,267 @@
...
+        echo "Generating nvidia-linux-headers..."
+        cp -av /oot-src/nvidia-oot/include /oot-src/out/nvidia-linux-header/
+        cat /src/Module.symvers \
+          /oot-src/out/nvidia-linux-header/Module.symvers.nvidia 2>/dev/null \
+          > /oot-src/out/nvidia-linux-header/Module.symvers
+      - |
+        echo "Building nvgpu module (OE4T patches-r36.5)..."
+        mkdir -p /usr/local/bin
+        printf '#!/bin/bash\nfiltered=()\nfor arg in "$@"; do\n  case "$arg" in\n    -fmin-function-alignment=*|-fconserve-stack) ;;\n    -fsanitize=bounds-strict) filtered+=("-fsanitize=bounds") ;;\n    -Wimplicit-fallthrough=*) filtered+=("-Wimplicit-fallthrough") ;;\n    -Wno-maybe-uninitialized) filtered+=("-Wno-uninitialized") ;;\n    -Wno-alloc-size-larger-than|-Wno-alloc-size-larger-than=*) ;;\n    -fplugin=*|-fplugin-arg-*) ;;\n    -pg|-mrecord-mcount|-mfentry|-fpatchable-function-entry=*) ;;\n    *) filtered+=("$arg") ;;\n  esac\ndone\nexec clang "${filtered[@]}" -Wno-unknown-warning-option -Wno-enum-enum-conversion -Wno-implicit-fallthrough -Wno-gnu-variable-sized-type-not-at-end\n' > /usr/local/bin/clang-oot
+        chmod +x /usr/local/bin/clang-oot
```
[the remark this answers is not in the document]
schwankner: Fixed — removed the duplicate creation from the nvgpu build block. The wrapper is installed once in the preceding OOT build step (`/usr/local/bin/clang-oot`) and persists on disk across shell blocks, so the nvgpu step now just references it directly.

nvidia-tegra-nvgpu/nvhost_ctrl_shim.c
```diff
@@ -0,0 +1,717 @@
...
+	if (!np) {
+		pr_err_ratelimited("nvhost-ctrl-shim: no host1x OF node found\n");
+		return ERR_PTR(-ENODEV);
+	}
+
+	pdev = of_find_device_by_node(np);
+	of_node_put(np);
+	if (!pdev) {
+		pr_err_ratelimited("nvhost-ctrl-shim: no host1x platform_device\n");
+		return ERR_PTR(-EAGAIN);
```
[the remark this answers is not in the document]
schwankner: Fixed — `put_device(&pdev->dev)` is now called immediately after `platform_get_drvdata()`, before the NULL check. This releases the reference on both the success and the error path. `drvdata` (the `struct host1x *`) remains valid independently of `pdev` as long as `host1x.ko` is loaded.

nvidia-tegra-nvgpu/nvhost_ctrl_shim.c
```diff
@@ -0,0 +1,717 @@
...
+	}
+
+	drvdata = platform_get_drvdata(pdev);
+	if (!drvdata) {
+		pr_err_ratelimited("nvhost-ctrl-shim: host1x drvdata is NULL\n");
+		return ERR_PTR(-EAGAIN);
+	}
+
```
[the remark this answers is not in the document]
schwankner: Addressed by the same fix: `put_device(&pdev->dev)` is called unconditionally right after `platform_get_drvdata()`, so the error path that returns `ERR_PTR(-EAGAIN)` when `drvdata` is NULL also releases the reference correctly.

nvidia-tegra-nvgpu/nvhost_ctrl_shim.c
```diff
@@ -0,0 +1,717 @@
...
+		pr_warn_ratelimited("nvhost-ctrl-shim: unknown ioctl cmd=0x%08x\n", cmd);
+		return -ENOTTY;
+	}
+}
+
+// ── Device class / cdev setup ─────────────────────────────────────────────────
+
+static char *nvhost_ctrl_devnode(const struct device *dev, umode_t *mode)
+{
+	*mode = 0666;
```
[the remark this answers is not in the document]
schwankner: Fixed — added a `if (mode)` guard before dereferencing. Kept `0666`: the CUDA runtime (`libnvrm_host1x.so`) opens this device as an unprivileged container user; Talos Linux runs a single-workload model so world-readable GPU control is the expected configuration (same as NVIDIA's own `/dev/nvidia*` devices). Added a comment to document the rationale.

nvidia-tegra-nvgpu/pkg.yaml
```diff
@@ -0,0 +1,267 @@
...
+        printf 'softdep nvhost-ctrl-shim pre: host1x host1x-fence\n' \
+          > /rootfs/usr/lib/modprobe.d/nvidia-tegra.conf
+        printf 'softdep nvgpu pre: host1x nvmap host1x-fence nvhost-ctrl-shim mc-utils\n' \
+          >> /rootfs/usr/lib/modprobe.d/nvidia-tegra.conf
+
+        echo "Installed modules:"
+        find /rootfs -name "*.ko" | sort
+finalize:
+  - from: /rootfs
+    to: /rootfs
```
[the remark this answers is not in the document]
schwankner: Fixed — changed to `to: /` to match the convention used by every other kernel module package in this repo (gasket-driver, zfs, hailort, xdma-driver, etc.).

nvidia-tegra-nvgpu/pkg.yaml
```diff
@@ -0,0 +1,267 @@
...
+        find /oot-src/nvidia-oot/drivers/gpu/host1x -name "host1x.ko" \
+          -exec install -m 644 {} /rootfs/usr/lib/modules/${KERNEL_RELEASE}/kernel/drivers/gpu/host1x/ \;
+
+        # host1x-fence + nvhost-ctrl-shim + all other nvidia-oot modules go to extra/
+        find /oot-src/nvidia-oot/drivers/gpu/host1x-fence -name "*.ko" \
+          -exec install -m 644 {} /rootfs/usr/lib/modules/${KERNEL_RELEASE}/extra/nvidia-tegra/ \;
+        find /oot-src/nvhost-ctrl-shim -name "*.ko" \
+          -exec install -m 644 {} /rootfs/usr/lib/modules/${KERNEL_RELEASE}/extra/nvidia-tegra/ \;
+        find /oot-src/nvidia-oot -name "*.ko" \
+          -exec install -m 644 {} /rootfs/usr/lib/modules/${KERNEL_RELEASE}/extra/nvidia-tegra/ \;
```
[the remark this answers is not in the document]
schwankner: Fixed — switched from `find /oot-src/nvidia-oot -name "*.ko"` (which caught the entire tree including host1x) to per-directory `modules_install` calls with explicit `INSTALL_MOD_DIR`. host1x now only installs to `kernel/drivers/gpu/host1x/` (shadow path) and all other modules only to `extra/nvidia-tegra/`, with no duplicates.

nvidia-tegra-nvgpu/pkg.yaml
```diff
@@ -0,0 +1,348 @@
+name: nvidia-tegra-nvgpu
+variant: scratch
+shell: /bin/bash
+dependencies:
+  - stage: base
+  - stage: kernel-build
+  - image: "{{ .LLVM_IMAGE }}:{{ .TOOLS_REV }}"
+steps:
```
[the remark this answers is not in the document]
frezbo: also our convention is name: `nvidia-tegra-nvgpu-pkg`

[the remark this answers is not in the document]
schwankner: Ok, I've made the necessary changes in commit [311f187](https://github.com/siderolabs/pkgs/pull/1518/commits/311f1870f16ad4e509c8e1c2b55581a5344f683b)

nvidia-tegra-nvgpu/pkg.yaml
```diff
@@ -0,0 +1,348 @@
...
+        NETLIST_PRIV=/oot-src/nvgpu/drivers/gpu/nvgpu/common/netlist/netlist_priv.h
+        if [ -f "${NETLIST_PRIV}" ]; then
+          sed -i 's/struct netlist_region regions\[1\]/struct netlist_region regions[]/' "${NETLIST_PRIV}"
+        fi
+        sed -i '/CONFIG_GCC_PLUGIN_LATENT_ENTROPY/d' /src/include/config/auto.conf 2>/dev/null || true
+        sed -i '/CONFIG_GCC_PLUGIN_LATENT_ENTROPY/d' /src/include/generated/autoconf.h 2>/dev/null || true
```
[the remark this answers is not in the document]
schwankner: solved with [9e04f9c](https://github.com/siderolabs/pkgs/pull/1518/commits/9e04f9cfb73dc25f6e3c4425ac9ad622ddbc34f2)

nvidia-tegra-nvgpu/scripts/fixup.sh
```diff
@@ -0,0 +1,127 @@
+#!/bin/bash
+# Patch OOT module Makefiles: remove -Werror and add required include paths.
+# srctree.nvconftest and srctree.nvidia-oot are passed as make vars at build time.
+set -euo pipefail
+
+NVIDIA_OOT=/oot-src/nvidia-oot
+CONFTEST_OUT=/oot-src/out/nvidia-conftest
```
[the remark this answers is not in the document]
schwankner: solved with [6ea1e15](https://github.com/siderolabs/pkgs/pull/1518/commits/6ea1e15c77bebc8bb70580c133ef34e23c4c50a1)

nvidia-tegra-nvgpu/pkg.yaml
```diff
@@ -0,0 +1,348 @@
...
+          KBUILD_MODPOST_WARN=1 \
+          modules
+        echo "nvgpu done."
+    install:
+      - |
+        NVIDIA_OOT=/oot-src/nvidia-oot
+        NVIDIA_CONFTEST=/oot-src/out/nvidia-conftest
+        SHIM_DIR=/oot-src/nvhost-ctrl-shim
+        echo "Installing kernel modules..."
+
```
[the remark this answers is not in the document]
schwankner: solved with [08a6cfd](https://github.com/siderolabs/pkgs/pull/1518/commits/08a6cfdffc3d3c4afc7f8e4b5464228a1aeb48a3)

nvidia-tegra-nvgpu/scripts/fixup.sh
```diff
@@ -0,0 +1,126 @@
+#!/bin/bash
```
frezbo, line 1: do you think OOT accepts these as patches upstream?

Pkgfile
```diff
@@ -288,5 +288,19 @@ vars:
...
+  # renovate: datasource=git-refs versioning=git depName=https://github.com/OE4T/linux-nvgpu.git
```
frezbo, line 315: which ref is referenced, is this a tag?

We could just use:

```
❯ git ls-remote git://nv-tegra.nvidia.com/linux-nvgpu.git
```

Pkgfile
dsseng, line 1: Where do these revisions come from? They do not seem to be tags in the repos, so are they perhaps referenced by meta-tegra or whatever?

Please let us know how to recognize proper, released versions of these drivers and not WIP snapshots which may be unstable

nvidia-tegra-nvgpu/pkg.yaml
```diff
@@ -0,0 +1,354 @@
...
+        # Module load order: host1x → host1x-fence → nvhost-ctrl-shim → nvmap → nvgpu
+        mkdir -p /rootfs/usr/lib/modprobe.d
+        printf 'softdep nvhost-ctrl-shim pre: host1x host1x-fence\n' \
+          > /rootfs/usr/lib/modprobe.d/nvidia-tegra.conf
+        printf 'softdep nvgpu pre: host1x nvmap host1x-fence nvhost-ctrl-shim mc-utils\n' \
+          >> /rootfs/usr/lib/modprobe.d/nvidia-tegra.conf
+
+        echo "Installed modules:"
+        find /rootfs -name "*.ko" | sort
```
dsseng: Not sure if this is useful, we don't use anything like this anywhere, you can generally save the output to a local directory and inspect it there

nvidia-tegra-nvgpu/scripts/patch_nvhost_host1x.py
dsseng, line 1: why is it patched via a custom script and not a simple patch applied with the `patch` tool?

Also, is this issue documented and tracked somewhere like OE4T or other communities? Are you working on upstreaming and getting the patch reviewed?

Searching for `ERRATA_SYNCPT_INVALID_ID_0` yields no results

nvidia-tegra-nvgpu/nvhost_ctrl_shim.c
dsseng, line 1: If this module is used by CUDA libraries, why is this feature not supported in the driver itself? Do you plan to request reviews and upstream it?

nvidia-tegra-nvgpu/pkg.yaml
```diff
@@ -0,0 +1,354 @@
...
+          -f /oot-src/out/nvidia-conftest/nvidia/Makefile
+        echo "conftest done."
+      - |
+        /pkg/scripts/fixup.sh
+      - |
+        # ── Cross-compiler mismatch fix: Clang wrapper ────────────────────────────
+        # Strips GCC-only flags before passing to clang (source-independent fix).
+        mkdir -p /usr/local/bin
+        printf '#!/bin/bash\nfiltered=()\nfor arg in "$@"; do\n  case "$arg" in\n    -fmin-function-alignment=*|-fconserve-stack) ;;\n    -fsanitize=bounds-strict) filtered+=("-fsanitize=bounds") ;;\n    -Wimplicit-fallthrough=*) filtered+=("-Wimplicit-fallthrough") ;;\n    -Wno-maybe-uninitialized) filtered+=("-Wno-uninitialized") ;;\n    -Wno-alloc-size-larger-than|-Wno-alloc-size-larger-than=*) ;;\n    -fplugin=*|-fplugin-arg-*) ;;\n    -pg|-mrecord-mcount|-mfentry|-fpatchable-function-entry=*) ;;\n    *) filtered+=("$arg") ;;\n  esac\ndone\nexec clang "${filtered[@]}" -Wno-unknown-warning-option -Wno-enum-enum-conversion -Wno-implicit-fallthrough -Wno-gnu-variable-sized-type-not-at-end\n' > /usr/local/bin/clang-oot
+        chmod +x /usr/local/bin/clang-oot
```
dsseng: Could you please consider moving this script to a file in the repo and copying it in? Packing it in a single printf is unnecessary and complicates reading

nvidia-tegra-nvgpu/pkg.yaml
```diff
@@ -0,0 +1,354 @@
...
+          INSTALL_MOD_STRIP=1 \
+          CONFIG_MODULE_SIG_ALL=y \
+          modules_install
+
+        # Module load order: host1x → host1x-fence → nvhost-ctrl-shim → nvmap → nvgpu
+        mkdir -p /rootfs/usr/lib/modprobe.d
+        printf 'softdep nvhost-ctrl-shim pre: host1x host1x-fence\n' \
+          > /rootfs/usr/lib/modprobe.d/nvidia-tegra.conf
+        printf 'softdep nvgpu pre: host1x nvmap host1x-fence nvhost-ctrl-shim mc-utils\n' \
+          >> /rootfs/usr/lib/modprobe.d/nvidia-tegra.conf
```
dsseng: Most if not all modules do not require this. I guess `depmod` in the imager should take care of configuring dependencies, but not sure how it's handled in general. Have you tried solving the issue in another way than manually creating these dependencies?

cc @frezbo

nvidia-tegra-nvgpu/scripts/fixup.sh
dsseng: Yes, ideally this should be handled on their side, rather than patching files that have been generated by from Kconfig.

nvidia-tegra-nvgpu/pkg.yaml
frezbo: ah yes, i forgot about this, imager should handle it, no point to have here

nvidia-tegra-nvgpu/scripts/fixup.sh
schwankner: The `kref_init(&syncpt[0].ref)` fix in `fixup.sh` is already present in newer OE4T commits (e.g., `6e071c0` in `linux-nv-oot`) — bumping the `oe4t_nv_oot_commit` pin would make that fixup redundant.

The `nvhost_host1x.c` retry loop addresses a real hardware behavior (`NVGPU_ERRATA_SYNCPT_INVALID_ID_0`, defined in the nvgpu errata table). I haven't submitted it upstream yet.

The `nvhost-ctrl-shim` module would require NVIDIA or OE4T to accept a `/dev/nvhost-ctrl` ABI compatibility layer for mainline-based kernels.

Pkgfile
schwankner: Valid concern. The three commits come from:
- `oe4t_nvgpu_commit` (d530a48): OE4T `linux-nvgpu` branch `patches-rel-36` — the project's default stable branch for the r36.x series
- `oe4t_nv_oot_commit` (ccf7646): OE4T `linux-nv-oot` branch `wip-r36.5-take-2` — admittedly a WIP branch. A more stable alternative is `patches-r36.5` (last updated 2026-03-05 with kernel 6.18 build fixes). I can switch to that.
- `oe4t_hwpm_commit` (4d8a699): OE4T `linux-hwpm` default branch

None have release tags — that is the current state of OE4T for GA10B on kernel 6.x.

nvidia-tegra-nvgpu/scripts/patch_nvhost_host1x.py
schwankner: On tooling: I can convert the source modifications to `.patch` files if preferred — the two code changes (`nvhost_host1x.c` retry loop, `netlist_priv.h` flexible array fix) are straightforward to express that way. The Makefile additions in `fixup.sh` are append-only so `sed` is conventional there, but I can switch if preferred.

On the errata name: `NVGPU_ERRATA_SYNCPT_INVALID_ID_0` is defined in the nvgpu source at [`drivers/gpu/nvgpu/include/nvgpu/errata.h`](https://github.com/OE4T/linux-nvgpu/blob/patches-rel-36/drivers/gpu/nvgpu/include/nvgpu/errata.h) in the OE4T repository. It is checked in `gk20a_channel_alloc_syncpt()` and causes channel init to fail when syncpt id=0 is returned.

nvidia-tegra-nvgpu/nvhost_ctrl_shim.c
schwankner: `/dev/nvhost-ctrl` is part of NVIDIA's traditional downstream Tegra kernel ABI (`nvidia-t23x-kernel`). On a standard JetPack Linux install, this device is provided by NVIDIA's downstream kernel. The OOT nvgpu driver provides GPU compute on mainline kernels but does not include the nvhost control interface.

On Talos Linux (mainline kernel + OOT modules), `libnvrm_host1x.so` cannot find `/dev/nvhost-ctrl` and falls back to CPU semaphore polling — which reduces LLM inference throughput from ~60 tok/s to ~7 tok/s on an Orin NX 16GB.

The long-term fix would be for NVIDIA to include nvhost-ctrl in the OOT driver or to update `libnvrm_host1x.so` to use a different interface. Until then, this shim is the only known way to enable hardware-interrupt-driven sync on mainline kernels.

nvidia-tegra-nvgpu/pkg.yaml
schwankner: Good point — I'll move the wrapper to `scripts/clang-oot` in the package directory and install it with `cp /pkg/scripts/clang-oot /usr/local/bin/clang-oot`.

nvidia-tegra-nvgpu/pkg.yaml
schwankner: Agreed — I'll remove the `modprobe.d/nvidia-tegra.conf` file. The Talos imager handles module load ordering via `depmod`.

nvidia-tegra-nvgpu/pkg.yaml
schwankner: Agreed — I'll remove that block.

nvidia-tegra-nvgpu/scripts/fixup.sh
schwankner: Agreed in principle. The `CONFIG_GCC_PLUGIN_LATENT_ENTROPY` removal is needed because the Talos kernel config (built for GCC) includes this option, but OOT modules are compiled with Clang which does not implement the GCC plugin — causing a compile error via `linux/random.h`. The correct fix is upstream: either in OE4T's build system or in Talos's kernel config for Clang builds. I'll track this as a follow-up.

Pkgfile
schwankner: The commits reference branch HEADs, not tags — OE4T does not publish release tags for these module trees. `nv-tegra.nvidia.com/linux-nvgpu.git` is publicly accessible but its latest branch is `l4t/l4t-r36.5` (JetPack 6.x, kernel 6.1) — there are no kernel 6.18 branches. The OE4T forks at `github.com/OE4T` are the only public source with kernel 6.18 support for Tegra234/GA10B.

That said, I can switch `oe4t_nv_oot_commit` from `wip-r36.5-take-2` to `patches-r36.5` (latest: `ea32e7f`, 2026-03-05) for a more stable reference. `oe4t_nvgpu_commit` is already on `patches-rel-36`, the default branch of `linux-nvgpu`.

schwankner: Done in [3f18915](https://github.com/siderolabs/pkgs/pull/1518/commits/3f189157a01f6c248391a3a36d505fe60cf39566).

nvidia-tegra-nvgpu/pkg.yaml
schwankner: Done in [6368e07](https://github.com/siderolabs/pkgs/pull/1518/commits/6368e07f63280f1c504e1ac6eea1de8bde04c1a6).

nvidia-tegra-nvgpu/pkg.yaml
schwankner: Done in [3624f1f](https://github.com/siderolabs/pkgs/pull/1518/commits/3624f1f4a53487c689590dceb781abe4a18e47fb) (removes both the modprobe.d file and the debug find output).

nvidia-tegra-nvgpu/nvhost_ctrl_shim.c
dsseng: Thank you for the explanation. I'm not very familiar with the ecosystem, but perhaps this solution is acceptable (with corresponding extension tier for a downstream driver).

Some public interaction with OE4T would be nice to get this driver reviewed by professionals of this platform, however

nvidia-tegra-nvgpu/scripts/fixup.sh
dsseng: Talos kernel is built with Clang now, and therefore modules are also. But perhaps autoconfiguration scripts in OE4T do not have proper support yet, right. Patching them to honor LLVM=1 (or passing the right parameters, maybe `CC`) would be a more robust solution

nvidia-tegra-nvgpu/scripts/patch_nvhost_host1x.py
dsseng: Yes, .patch would be preferred as the standard way we use for other code patches in the repo. You should be able to use `git diff` to derive them after you applied patches with the script.

I guess sed is okay for now for Makefile, and work should instead be focused on making the automatic config not need such fixups

schwankner: Done in [559098f](https://github.com/siderolabs/pkgs/pull/1518/commits/559098f). The Python script and the awk one-liner are replaced by two standard `.patch` files under `patches/` applied with `patch -p1`:

- `0001-nvgpu-syncpt-retry-errata-id0.patch` — retry loop + syncpt id=0 guard for GA10B `ERRATA_SYNCPT_INVALID_ID_0` (also adds the missing `#include <linux/delay.h>`)
- `0002-nvgpu-netlist-flexible-array.patch` — flexible array fix (`regions[1]` → `regions[]`)

The Makefile additions in `fixup.sh` remain as `sed` (append-only, no context lines to match).

nvidia-tegra-nvgpu/nvhost_ctrl_shim.c
schwankner: Good point. I'll open an issue in `OE4T/meta-tegra` to start the conversation — the key question for them is whether `/dev/nvhost-ctrl` should be integrated into the OOT nvgpu package for mainline deployments, or whether there's a different intended path for JetPack 6 CUDA on non-JetPack kernels. Will link the issue here once open.

nvidia-tegra-nvgpu/scripts/fixup.sh
schwankner: Agreed. The two categories in `fixup.sh` have different upstream paths:

1. **`CONFIG_GCC_PLUGIN_LATENT_ENTROPY` strip** — this is a Talos kernel config artifact: `CONFIG_GCC_PLUGIN_LATENT_ENTROPY=y` was set when `config-arm64` was generated with GCC. Since Talos now builds with Clang, removing it from `config-arm64` is the correct fix. I can open a separate PR in this repo for that.

2. **Conftest macro overrides** (`NV_IOMMU_MAP_HAS_GFP_ARG`, `NV_MM_STRUCT_...`, etc.) — these compensate for OE4T's conftest infrastructure not yet probing kernel 6.18 APIs correctly when building with Clang/LLVM=1. The proper fix lives in `OE4T/linux-nv-oot`: the conftest scripts need kernel 6.18 API detections. I'll track this in an OE4T meta-tegra issue (since `linux-nv-oot` has issues disabled). Once upstream conftest detections are fixed, these overrides can be removed.

nvidia-tegra-nvgpu/nvhost_ctrl_shim.c
schwankner: OE4T issue opened: https://github.com/OE4T/meta-tegra/issues/2196

frezbo: Since we're early in the 1.14 works, I would wait for a while and see what the OE4T folks have to say on this issue, let's wait a week or two before proceeding, does that sound good? Meanwhile we can have draft PR's ready for extensions/overlays

schwankner: Hi @frezbo, I got the OE4T feedback from the [disscussion](https://github.com/orgs/OE4T/discussions/2198).

Short summary: The maintainers position is that `/dev/nvhost-ctrl` was an older interface and they can't give a definitive answer on whether the shim approach is correct — since it involves NVIDIA's closed-source libraries. They recommended asking the NVIDIA developer forum.

While waiting, I did some deeper analysis on my running system. Key finding: the CUDA process holds both `/dev/nvgpu/igpu0/ctrl` (channel allocation) and `/dev/nvhost-ctrl` (syncpoint wait) open simultaneously during inference, not as a fallback sequence. And no module in the OOT stack creates the global `/dev/nvhost-ctrl` device. `host1x-nvhost.ko` only creates per-engine devices like `nvhost-ctrl-host1x`. So the shim fills a structural gap.

I will ask in the NVIDIA dev forum and then come back here.

I will wait to start on the extension/overlay draft PR until I have an answer there.

frezbo: great news, I also quickly glanced over the comments, and i think some of the comments makes sense, sometimes if both are present it might try to keep the old one too, LLM's might not catch it unless you prompt it to say it's old, that might be worth giving a try, I would suggest to pipe the conversation from OE4T into an LLM and see if it re-thinks

Anyways we still have some time to polish this up before 1.14

nvidia-tegra-nvgpu/scripts/fixup.sh
```diff
@@ -0,0 +1,99 @@
...
```
schwankner, line 1: Done — opened [siderolabs/pkgs#1529](https://github.com/siderolabs/pkgs/pull/1529) which explicitly disables `CONFIG_GCC_PLUGINS` and `CONFIG_GCC_PLUGIN_LATENT_ENTROPY` in `config-arm64`. The corresponding `sed` workaround in `fixup.sh` has been removed in this PR (commit 754aee5).

nvidia-tegra-nvgpu/scripts/fixup.sh
schwankner: Update: the dead code (CONFIG_GCC_PLUGIN_LATENT_ENTROPY removal + kref_init patch) has been removed from fixup.sh in commit [754aee5](https://github.com/schwankner/pkgs/commit/754aee57).

You're right that patching Kconfig-generated files is the wrong approach. A separate PR (#1529) to document the config-arm64 state was opened but then closed per your feedback — adding `# ... is not set` lines doesn't survive `make olddefconfig`.

The real reason this was dead code: with `LLVM=1` / `CC_IS_CLANG=y`, `CONFIG_GCC_PLUGINS` is never set to `=y` by Kconfig in the first place, so the sed commands never matched anything. Build-tested with nvgpu 5.10.9 so no regression.

schwankner: **Update on the conftest macro overrides (`#if 1` guards):**

Following up on @dsseng's suggestion to "patch them to honor LLVM=1": we experimented with replacing the `#if 1 /* force: kernel 6.x+ */` workarounds with `#if LINUX_VERSION_CODE >= KERNEL_VERSION(x,y,z)` guards (branch `feat/linux-version-code-guards`), but this is not viable in the siderolabs OOT build context:

- `linux/version.h` is **not present** in the kernel build container's `/src` include tree — that file is only generated during a full kernel build; the module-build headers package does not include it
- `KBUILD_CFLAGS` includes `-Wundef` on kernel 6.18, so an undefined `KERNEL_VERSION()` function-like macro is a fatal compile error
- Adding `-include $(srctree)/include/linux/version.h` to `ccflags-y` fails because the file doesn't exist at `/src/include/linux/version.h`

The `#if 1` guards remain the correct workaround for this build context.

**Root cause** of why the conftest probes fail in the first place: `NV_CONFTEST_CFLAGS` in OE4T's `scripts/conftest/Makefile` hardcodes `-Werror`. When Clang emits any warning during a conftest probe (Clang is stricter than GCC on signed/unsigned comparisons, implicit fallthrough, etc.), the probe fails → the feature macro is left undefined → the wrong code path is selected. The proper fix is in OE4T's conftest infrastructure.

We opened [OE4T/meta-tegra discussion #2198](https://github.com/OE4T/meta-tegra/discussions/2198) to track this, but the thread concluded without a fix commitment (OE4T treats it as a closed-source CUDA libs concern). The `#if 1` guards will stay until OE4T's conftest handles LLVM=1 + kernel 6.18 correctly.

dsseng: > linux/version.h is not present in the kernel build container's /src include tree — that file is only generated during a full kernel build; the module-build headers package does not include it

Hm, I'm not so sure about this, I tend to believe modules actually do have access to kernel version, at least some like DRBD explicitly verify kernel version compatibility. Try using `BUILDX_EXPERIMENTAL=1 docker buildx debug` + a `false` in one of the build steps to get a shell inside of the build sandbox, and you should be able to search for the files available to the module build

schwankner: you were right, linux/version.h is available, just not at the path I tried but at `/src/include/generated/uapi/linux/version.h`. Is this the file you mean?

Adding `-include $(srctree)/include/generated/uapi/linux/version.h` to `ccflags-y` makes `LINUX_VERSION_CODE` and `KERNEL_VERSION()` available at OOT module compile time. All `#if 1 /* force: kernel X+ */` hacks are now replaced with proper `#if LINUX_VERSION_CODE >= KERNEL_VERSION(x,y,z)` guards. One exception remains: `NV_DEVM_TEGRA_CORE_DEV_INIT_OPP_TABLE_COMMON_PRESENT` is an OE4T-internal function with no corresponding kernel version.

dsseng: I am not sure on this one, sorry. But certainly better to have less hardcoded things

schwankner:
Finally managed to post a question to the NVIDIA Developer Forum to get clarity on the `/dev/nvhost-ctrl` situation: [/dev/nvhost-ctrl missing with JetPack 6 Out-of-Tree modules on mainline kernel](https://forums.developer.nvidia.com/t/dev-nvhost-ctrl-missing-with-jetpack-6-out-of-tree-modules-on-mainline-kernel/370627)

frezbo:
> Finally managed to post a question to the NVIDIA Developer Forum to get clarity on the `/dev/nvhost-ctrl` situation: [/dev/nvhost-ctrl missing with JetPack 6 Out-of-Tree modules on mainline kernel](https://forums.developer.nvidia.com/t/dev-nvhost-ctrl-missing-with-jetpack-6-out-of-tree-modules-on-mainline-kernel/370627)

They seemed to ask more questions to try with stock one

mmalyska:
Hi @schwankner Any updates on this?

schwankner:
Hi @mmalyska, @frezbo. Sorry for the long delay, I've been very busy privately and hadn't been able to get back to testing until now. It's tested now.

Short version of what the actual problem was: the blocker was never `/dev/nvhost-ctrl`. The stock mainline `tegra-drm.ko` is built against the upstream `host1x` ABI and refuses to load on top of the OE4T-patched `host1x.ko` (unresolved `host1x_job_alloc`, `err -22`), so `/dev/dri/renderD128` never appeared and `cuInit` returned 801. NVIDIA confirmed in the [forum thread](https://forums.developer.nvidia.com/t/dev-nvhost-ctrl-missing-with-jetpack-6-out-of-tree-modules-on-mainline-kernel/370627) that stock JetPack 6 also has no generic `/dev/nvhost-ctrl` (only `nvhost-ctrl-gpu`).

The fix is to build **`tegra-drm`, `host1x-nvhost` and `nvhwpm` from the same OE4T tree** as `host1x` so the ABI matches. With the DRM render node present, CUDA works **without the shim at all**.

I've verified this on a fresh Jetson Orin NX 16 GB (Talos v1.13.0, kernel 6.18.24), confirmed real GPU usage (models fully in VRAM, `library=CUDA`), ~64 tok/s (qwen2.5:0.5b) / ~16 tok/s (qwen3:4b), on par with native JetPack 6.2.

Given this, **I will update this PR accordingly**, dropping `nvhost_ctrl_shim` and switching to the OE4T DRM stack, since that's the approach that actually works. I'll refresh the branch accordingly. Reference implementation: https://github.com/schwankner/talos-jetson-orin

schwankner:
Pushed the update. The shim is gone and the package now builds the OE4T DRM stack instead:

- `host1x-nvhost`, `nvhwpm` (OE4T linux-hwpm) and `tegra-drm` are built from the same OE4T tree as `host1x`, so the ABI matches
- `tegra-drm` is installed at `kernel/drivers/gpu/drm/tegra/` so depmod/modprobe resolve this module rather than the in-tree one, which is what caused the `host1x_job_alloc` version mismatch
- `tegra-drm` is built headless (`CONFIG_DRM_TEGRA_HAVE_DISPLAY=n`); only the render node is needed
- load order is now `host1x -> host1x-fence -> host1x-nvhost -> tegra-drm -> nvmap -> nvgpu`
- `nvhost_ctrl_shim.c` removed (-510 lines)

Two build fixes came out of validating this, both pre-existing:

1. `scripts/fixup.sh` and `scripts/clang-oot` were committed without the executable bit, so `/pkg/scripts/fixup.sh` aborted with exit code 126. Mode is set now and the script is invoked through `bash` so it no longer depends on the mode.
2. `CONFIG_GCC_PLUGIN_LATENT_ENTROPY` is now stripped before building. It declares a global via a GCC plugin that Clang does not run, which breaks `linux/random.h` for every OOT module. The sed is a no-op when the option is not set.

Since a fork cannot use the `pkgs` runner group, I validated the build separately on an arm64 runner: all modules compile (`host1x`, `host1x-fence`, `host1x-nvhost`, `nvhwpm`, `tegra-drm`, `nvmap`, `mc-utils`, `nvgpu`), no shim is produced, and both `host1x` and `tegra-drm` land at their shadow paths. Runtime verification was done on a Jetson Orin NX 16 GB as described above.

Happy to split the two build fixes into a separate commit or rebase if you prefer, and to add other Orin variants to the description.

schwankner:
Rebased onto current main to resolve a merge conflict (siderolabs/pkgs main had moved 122 commits ahead, including another PR adding gds-nvidia-fs vars at the same insertion point in Pkgfile as this one's oe4t_* vars; both blocks are independent and now sit side by side).

Re-validated against the latest main (kernel 6.18.45): all modules build (host1x, host1x-fence, host1x-nvhost, nvhwpm, tegra-drm, nvmap, mc-utils, nvgpu), no shim, both host1x and tegra-drm at their shadow paths. No changes were needed in nvidia-tegra-nvgpu/ itself, only the Pkgfile conflict.

nvidia-tegra-nvgpu/scripts/clang-oot
frezbo, line 1: maybe env vars can be used to override?

frezbo: I meant CFLAGS etc

nvidia-tegra-nvgpu/scripts/fixup.sh
```diff
@@ -0,0 +1,126 @@
...
+# srctree.nvconftest and srctree.nvidia-oot are passed as make vars at build time.
+set -euo pipefail
+
+NVIDIA_OOT=/oot-src/nvidia-oot
+
+# CONFIG_GCC_PLUGIN_LATENT_ENTROPY injects a global 'latent_entropy' variable via
+# a GCC plugin. These modules are built with Clang, which does not run the plugin,
+# so the symbol is undeclared and linux/random.h fails to compile for every OOT
+# module. Drop the macro before any module build. No-op when it is not set.
+sed -i '/CONFIG_GCC_PLUGIN_LATENT_ENTROPY/d' /src/include/config/auto.conf 2>/dev/null || true
```
frezbo: is there anyting pending that can be simplified in here, this is still huge, maybe a patch file is easier to maintain

schwankner:
Thanks for the review, and sorry for the slow turnaround. Both suggestions were right, and
digging into them turned up the actual reason the recipe had grown so large. I pushed the
result: pkg.yaml is down from 547 to 137 lines, `scripts/fixup.sh` and `scripts/clang-oot`
are gone.

**The CC wrapper is dead code.** I instrumented it to log every flag it touched. Across more
than 1500 compiler invocations it did exactly two things: it dropped
`-fpatchable-function-entry=4,2`, which Clang supports and ftrace needs, and it rewrote
`-Wimplicit-fallthrough=3` to `-Wimplicit-fallthrough`, which comes from nvgpu's own
`cc-option` probe that already drops the flag by itself under Clang. No GCC-only flags reach
the compiler any more now that the pkgs kernel is Clang-built, so the wrapper was at best
useless. The four `-Wno-*` flags it appended turned out to be unnecessary as well once the right code
paths are taken; the only global flags left are the include paths, passed as `KCFLAGS` in the
package `env:`, which is what you suggested.

**The macro rewrites had a root cause.** `scripts/conftest/Makefile` uses `CC ?= cc`, and `cc`
in the build image is GCC. Every probe was compiled with GCC against the headers of a
Clang-built kernel, so the probes failed wholesale and every `NV_*` macro stayed undefined.
The sources then took the wrong branch, for example host1x `cdma.c` calling `iommu_map()`
with five arguments. That is what the roughly 30 `LINUX_VERSION_CODE` rewrites were papering
over. Running conftest with `CC=clang` makes all probes behave, and the rewrites are gone.
The conftest Makefile also hardcodes `-Werror` in `NV_CONFTEST_CFLAGS`, so probes that merely
warn under Clang fail as well; I drop it on the working copy so a probe only fails on a real
compile error. The recipe now asserts that one macro every 6.3+ kernel must have is defined,
so a broken conftest fails the build instead of silently selecting old code paths.

**What is left.** Three patch files under `nvidia-tegra-nvgpu/patches/`, grouped per source tree:
the headless tegra-drm build (the kernel config has `DRM_FBDEV_EMULATION=y`, which the module
inherits through `autoconf.h`), the GA10B syncpoint retry and id-0 workaround, and a netlist
flexible-array change that silences a UBSAN false positive. Three further fixes I carried
(`class_create()` owner argument, the `devnode` signature, `get_user_pages()` without `vmas`)
turned out to be guarded by `NV_*` macros in the OE4T sources already, so they went away with
the working conftest. `-Werror` in the OE4T module Makefiles is filtered with
`ccflags-remove-y=-Werror` on the make command line instead of editing nine Makefiles, only the
make variables that differ from the OE4T defaults are passed, and one shell function covers all
nine module directories in both the build and install steps.

I validated this the same way as before, plus on hardware this time. The build links all ten
modules with no errors, and on a Jetson Orin NX the extension gives `/dev/dri/renderD128`,
`library=CUDA compute=8.7` in Ollama, and 66.0 tok/s on qwen2.5:0.5b against 65 to 66 tok/s
for the previous recipe, so the simplification is behaviour-neutral.

Two open points from my side: the GA10B firmware still has to come from somewhere, either
linux-firmware in pkgs or bundled in the extension, and there is no Jetson GPU extension in
siderolabs/extensions yet. Happy to follow whichever route you prefer.

dsseng:
> Two open points from my side: the GA10B firmware still has to come from somewhere, either
> linux-firmware in pkgs or bundled in the extension, and there is no Jetson GPU extension in
> siderolabs/extensions yet. Happy to follow whichever route you prefer.

`linux-firmware` in pkgs contains all the firmwares officially shipped by linux-firmware. If the needed blobs are available - please create the extension consisting of this driver and the necessary firmware files copied from there. If not, please find another clearly licensed and reliable source, and use that for the extension

schwankner:
Thanks, that settles the direction. I checked both routes:

**linux-firmware does not carry the GA10B blobs.** The upstream tree has `nvidia/gk20a`, `gm20b`, `gp10b`, `gv100`, the `tegra*` bootloader/XUSB files and the desktop GSP firmware (`ga100`, `ga102`, ...), but no `nvidia/ga10b`. NVIDIA has never submitted the Orin GPU firmware there, so the pkgs `linux-firmware` package cannot supply it.

**The clearly licensed source is NVIDIA's own Jetson Linux package.** The 17 files nvgpu loads on Orin ship in `nvidia-l4t-firmware` from the official Jetson apt repository (`repo.download.nvidia.com/jetson/t234`, currently `36.5.2-20260716114719`), under `lib/firmware/nvidia/ga10b/`. Its copyright file places them under the NVIDIA Driver License Agreement, and section 1.1(d) of that agreement explicitly permits distribution "for use with operating system kernels distributed under the terms of an OSI-approved open source license", provided the binaries stay unmodified and the agreement is shipped alongside. That is the same license text siderolabs/extensions already carries as `NVIDIA-driver-LICENSE` for the nonfree `kmod-nvidia` and Kata GPU extensions, so the packaging would follow the existing pattern rather than introduce a new one. OE4T's meta-tegra takes the firmware from the same deb for its Yocto BSP.

**Proposal.** I would add a nonfree extension to siderolabs/extensions next to `nvidia-gpu/nonfree/kmod-nvidia`, depending on the `nvidia-tegra-nvgpu-pkg` image from this PR and fetching the pinned `nvidia-l4t-firmware` deb (URL plus sha256 in vars) to copy exactly the `ga10b` directory into `/usr/lib/firmware/ga10b`, together with the license file and the `modprobe.d` softdep ordering. Firmware and modules come from the same JetPack release, so they stay in lockstep. Once this PR is merged I will open that extension PR; if you would rather see the extension first, I can open it against my fork's pkgs image so both can be reviewed together.

quom:
I tested this on an NVIDIA Jetson AGX Orin 64GB and successfully ran a suite of image models benchmarking performance using the GPU

PR #1528, opened 2026-05-01
fix: kernel hardening pack of changes

smira:
A pack of various kernel hardening efforts. 

This needs a proper test before moving forward.


kernel/build/config-amd64
```diff
@@ -313,7 +313,7 @@ CONFIG_KEXEC_CORE=y
 # CONFIG_KEXEC is not set
 CONFIG_KEXEC_FILE=y
 CONFIG_KEXEC_SIG=y
-# CONFIG_KEXEC_SIG_FORCE is not set
+CONFIG_KEXEC_SIG_FORCE=y
```
frezbo: this will break kexec with UKI on non-secureboot

smira: going to test this to ensure we have full picture!

smira: yeah, it breaks things, going to drop it.

smira:
Console seems to work still, verified UEFI mode.

smira:
> Console seems to work still, verified UEFI mode.

... verified BIOS console as well.

frezbo (approved):
🆒

smira:
The plan for further testing - merge it and test on main more through various pipelines.

PR #1529, opened 2026-05-03
fix: disable GCC plugins in config-arm64 (OOT Clang build compat)

schwankner:
## Problem

Talos arm64 kernel is built with Clang (`LLVM=1`), but `config-arm64` does not
explicitly disable GCC plugin options. With `CONFIG_HAVE_GCC_PLUGINS=y` in place,
Kconfig may propagate `CONFIG_GCC_PLUGIN_LATENT_ENTROPY=y` into
`include/generated/autoconf.h` if the config was generated under GCC, or if
toolchain detection runs before `CC_IS_CLANG` is established.

Out-of-tree modules that include kernel headers then fail at compile time:

```
include/linux/random.h:24:22: error: use of undeclared identifier 'latent_entropy'
```

because `latent_entropy` is a GCC plugin-injected variable that Clang never
defines.

## Fix

Explicitly disable both options in `config-arm64`:

```
# CONFIG_GCC_PLUGINS is not set
# CONFIG_GCC_PLUGIN_LATENT_ENTROPY is not set
```

This ensures OOT module builds against the Talos kernel headers succeed with
Clang regardless of how or when `autoconf.h` was generated.

## Context

Discovered during the nvidia-tegra-nvgpu OOT module build for Jetson Orin
(PR #1518). The nvidia-tegra-nvgpu `fixup.sh` previously worked around this by
stripping the option from `auto.conf`/`autoconf.h` at build time; this PR
provides the proper upstream fix so the workaround is no longer needed.

References: #1518


dsseng (commented):
Unfortunately, this does not fix the problem, because `make olddefconfig` (used on kernel updates, for example) will overwrite this change. These lines are never generated in first place because they are not disabled, but absent from the intermediate Kconfig, being behind an `if` block.

I think this should actually be fixed on the receiving side, i.e. look not for `is not set`, but rather for `=y` to find features that are certainly enabled

FYI - it's rarely the case PRs are opened directly to the `release-*` branch, the typical workflow is merging the fix on main and backporting it. But no need to switch this PR as it's likely to require a different fix

schwankner:
Thanks for the review @dsseng — you're correct. Adding `# CONFIG_GCC_PLUGIN_LATENT_ENTROPY is not set` to `config-arm64` does not survive `make olddefconfig`, since the entry lives behind an `if GCC_PLUGINS` Kconfig block and is simply absent (not "not set") in LLVM builds.

Closing this PR. The dead code in `fixup.sh` that prompted it has already been removed (commit [754aee5](https://github.com/schwankner/pkgs/commit/754aee5) on the main PR branch) — with `LLVM=1` / `CC_IS_CLANG=y`, GCC plugins are never active in the first place, so the sed commands were unreachable. Build-tested with nvgpu 5.10.9 ✅.

If a proper upstream fix is needed (OOT conftest checking for `=y` instead of `is not set`), that belongs in OE4T/linux-nv-oot.

PR #1566, opened 2026-05-24
fix: enable CONFIG_BCM2712_MIP as built-in in arm64 kernel config

appkins:
*IMPORTANT*: This change is **required** for working ethernet on Raspberry Pi 5.

This pull request updates the configuration for the ARM64 kernel build, specifically changing the build setting for the Broadcom BCM2712 MIP driver from a loadable module to being built directly into the kernel.

Kernel configuration update:

* Changed `CONFIG_BCM2712_MIP` from module (`m`) to built-in (`y`) in `kernel/build/config-arm64`, ensuring the BCM2712 MIP driver is always included in the kernel image.


smira:
What is the problem with module? Doesn't it get auto-loaded?

appkins:
> What is the problem with module? Doesn't it get auto-loaded?

It's the timing. When booting on rpi 5, Linux attempts to probe via rp1_pci before the module is loaded. This fails due to the missing MIP. It then fails to ever load the RP1 south bridge which means no GPIO and no ethernet. This change fixes the boot process and Talos runs as expected on RPI 5.

smira (approved):

PR #1578, opened 2026-06-08
fix: avoid page_table_check BUG on time namespace VVAR page

smira:
Fixes https://github.com/siderolabs/talos/issues/13496

## What

Adds a kernel patch (`0006-mm-page_table_check-do-not-track-special-PFN-mapped-PTEs.patch`) that stops `page_table_check` from tracking special (PFN-mapped) PTEs.

## Why

Reported in [siderolabs/talos#13496](https://github.com/siderolabs/talos/issues/13496): nodes panic + reboot under heavy container/CI churn with:

```
kernel BUG at mm/page_table_check.c:143!
__page_table_check_zero
__free_frozen_pages
free_time_ns
free_nsproxy
do_exit / do_group_exit
```

### Root cause

In 6.18.y the vDSO `[vvar]` mapping is `VM_PFNMAP` and its pages are installed into userspace via `vmf_insert_pfn()`, which produces **special** PTEs. On x86 and arm64 `pte_user_accessible_page()` only checks the PRESENT/USER bits and does **not** exclude `pte_special()`, so `page_table_check` accounts these PFN mappings in the per-page map counters — even though they are not rmap-managed (`vm_normal_page()` returns `NULL`).

Most vvar pages live in the kernel image and are never freed, so the stray accounting is invisible. The **time-namespace VVAR page** is the exception: it is a real `alloc_page()` page freed by `__free_page()` in `free_time_ns()` when the last task of a time namespace exits. The unbalanced special-PTE accounting leaves a non-zero `file_map_count`, which trips the `BUG_ON()` in `__page_table_check_zero()` at free time.

This is why:
- it is arch-independent (x86_64 + arm64 here, riscv via syzbot),
- only the timens page trips it (it's the only one in this set that is freed),
- it only affects Talos / kernels with `CONFIG_PAGE_TABLE_CHECK` enabled,
- it is triggered by `CLONE_NEWTIME` churn (runc / docker-init / tini),
- and `page_table_check=off` is a complete workaround.

### Fix

Skip special PTEs in both the set and clear paths of `page_table_check`, so the counters stay balanced (always zero) for PFN-mapped pages. Special PTEs have no struct-page rmap semantics and should never have been tracked. `pte_special()` is generically available (no-op `false` on arches without `ARCH_HAS_PTE_SPECIAL`), so it is a single arch-independent change.

[...]

PR #1580, opened 2026-06-08
fix: disable PAGE_TABLE_CHECK_ENFORCED in kernel config

shanduur:
Enforced check panics on kernel BUG at mm/page_table_check.c:143
during process cleanup, causing node reboots. Keep
CONFIG_PAGE_TABLE_CHECK=y so it stays off by default but opt-in
via page_table_check=on cmdline.

Refs siderolabs/talos#13496


smira (approved):

frezbo (approved):

Orzelius (commented):
would it be appropriate to add a comment that it should be re-enabled once we have the fixed kernel?

shanduur:
@Orzelius this targets only release 1.13 (and potentially 1.12) branch - we will have a `.patch` for `main` to test it before 1.14 release.

See https://github.com/siderolabs/pkgs/pull/1578

PR #1599, opened 2026-07-01
feat: add mdadm package for software RAID

shanduur:
Move mdadm from extensions into core packages.


mdadm/pkg.yaml
```diff
@@ -0,0 +1,44 @@
...
+    build:
+      - |
+        mkdir -p /run/mdadm
+
+        CXFLAGS=-D_LARGEFILE64_SOURCE \
+        make -j $(nproc) mdadm
+        make -j $(nproc) mdmon
+    install:
+      - |
+        mkdir -p /rootfs/usr/local/sbin
```
smira: should be `/usr/sbin` if we move it to core

mdadm/pkg.yaml
```diff
@@ -0,0 +1,44 @@
...
+        sha512: {{ .mdadm_sha512 }}
+    prepare:
+      - |
+        tar -xzf mdadm.tar.gz --strip-components=1
+      - |
+        patch -p1 < /pkg/patches/mdadm-include-sysmacros.patch
+        patch -p1 < /pkg/patches/no-werror.patch
+    build:
+      - |
+        mkdir -p /run/mdadm
```
smira, line 23: is this used? (what does it do?)

smira (approved):
we can't drop extensions, but we can make them empty (no-op) to keep existing schematics working on upgrade

mdadm/pkg.yaml
shanduur: Without this it will fail to compile:

```
0.270 ***** Parent of /run/mdadm does not exist.  Maybe set different RUN_DIR=
0.270 *****  e.g. make RUN_DIR=/dev/.mdadm
0.270 ***** or set CHECK_RUN_DIR=0
```

And setting `CHECK_RUN_DIR=0` does not work - it would require Makefile patch, as it is not overridable.

PR #1600, opened 2026-07-01
fix: use non-conflicting name for mdadm

shanduur:
`name: mdadm` was used by extension.
We need new name that does not conflict with it.


mdadm/pkg.yaml
```diff
@@ -1,4 +1,4 @@
-name: mdadm
+name: mdadm-pkg
```
smira, line 1: I think we need to update .kres.yaml -> Makefile

smira (approved):

PR #1607, opened 2026-07-10
feat: update DRBD to 9.3.3

dedo1911:
Backportable change.

Updates the DRBD kernel driver from 9.3.2 to the upstream 9.3.3 release (LINBIT/drbd).


smira (approved):
thank you, I also regenerated the compat files, I hope it builds.

Issue #1276, opened 2025-07-09
kernel: enable `CONFIG_AMD_MEM_ENCRYPT`


james-callahan:
https://github.com/siderolabs/pkgs/blob/c97d25e102cb7e3d9e295260490ef949fc7ea757/kernel/build/config-amd64#L455

This is required to turn on confidential compute on GCP: https://cloud.google.com/confidential-computing/confidential-vm/docs/create-custom-confidential-vm-images#amd-sev-and-sev-snp

<img width="881" height="550" alt="Image" src="https://github.com/user-attachments/assets/f24af266-630a-4d44-b734-21eb33fcf4f8" />

(the other config options are already turned on)

PR #1608, opened 2026-07-13
feat: enable CONFIG_IOMMUFD and CONFIG_VFIO_DEVICE_CDEV

calintkd:
Enables the VFIO IOMMUFD character-device interface
(`/dev/vfio/devices/vfioN`) so VFIO device passthrough into confidential VMs
works on the stock Talos kernel. The Kata Containers runtime refuses the
legacy VFIO group interface for confidential guests
(`src/runtime/virtcontainers/qemu.go`):

```
ConfidentialGuest needs IOMMUFD - cannot use /dev/vfio/<group>
```

This extends the Kata-on-Talos pattern — already shipped as the in-catalog
`kata-containers` extension (non-confidential today) — to confidential guests
with device passthrough, e.g. an NVIDIA GPU passed into an SEV-SNP guest,
continuing the confidential-computing enablement from #1396 (SEV-SNP) and
#1276 (AMD_MEM_ENCRYPT). Rebinding devices to `vfio-pci` is already
first-class in Talos via `PCIDriverRebindConfig`; for confidential guests the
kernel-side cdev interface is the missing piece.

### Config delta

`config-amd64` regenerated with `make kernel-olddefconfig` after setting
`CONFIG_IOMMUFD=y` + `CONFIG_VFIO_DEVICE_CDEV=y`; the net change is 5 symbols:
`VFIO_DEVICE_CDEV` and `IOMMUFD` (set), plus auto-selected `IOMMUFD_DRIVER`,
`IOMMUFD_DRIVER_CORE`, and `INTERVAL_TREE_SPAN_ITER`. `VFIO_GROUP=y` and the
legacy container path are unchanged, and `IOMMUFD_VFIO_CONTAINER` stays off
(it depends on `!VFIO_CONTAINER`). `/dev/iommu` is a root-only 0660 misc
device — stricter than the 0666 `/dev/vfio/vfio` already enabled — and the
build's `kernel-hardening-checker` gate still passes (neither it nor KSPP has
a rule against IOMMUFD).

### Why `=y` rather than `=m`

Tested both on Talos v1.13.5. With `=m`, `vfio.ko` is built with the iommufd
glue (`drivers/vfio/iommufd.c`) and hard-depends on iommufd's exported
symbols, but `iommufd.ko` is not in the Talos rootfs module allowlist
(`hack/modules-amd64.txt` carries the vfio stack, no iommufd entry) — so the
vfio modules fail to load with `Unknown symbol iommufd_*`. `=m` therefore
needs a companion siderolabs/talos allowlist change (after which depmod would
auto-load it), while `=y` is this single file. Happy to switch this PR to the
`=m` + allowlist pair if you prefer modular.

### Testing

Built this kernel and ran it on a Talos v1.13.5 bare-metal AMD SEV-SNP node
with an NVIDIA H200 passed through to a Kata confidential guest:
`/dev/vfio/devices/vfio0` present, guest boots, `nvidia-smi` works inside the
CVM. A stock kernel reproduces the `ConfidentialGuest needs IOMMUFD` failure.
Cost: +48 KiB compressed vmlinuz (+0.24%).

amd64 only — the validated confidential-VM stack is SEV-SNP; arm64 (SMMUv3
supports iommufd) can follow identically if you want parity. If this is
eligible for a `release-1.13` cherry-pick that would be welcome.

New module:

```
kernel/drivers/iommu/iommufd/iommufd.ko
```


smira (approved):
thank you, I moved it to use modules when possible

calintkd:
Thanks @smira — validated the `=m` build end-to-end on SEV-SNP + H200: with `iommufd.ko` in the rootfs allowlist it auto-loads as a `vfio` dependency and GPU passthrough into the confidential guest works.

PR #1612, opened 2026-07-14
chore: bump bldr to v0.6.1

majabojarska:
Makefile bldr version was already bumped in 3f09c0c8d9d5522d0f78a86df98971ae7913afea

### DIFF
--- a/Pkgfile
+++ b/Pkgfile
@@ -1,4 +1,4 @@
-# syntax = ghcr.io/siderolabs/bldr:v0.6.0
+# syntax = ghcr.io/siderolabs/bldr:v0.6.1
 
 format: v1alpha2
 


frezbo (approved):
I wonder if we should bring Pkgfile headers under kres control 🤔

shanduur (approved):

majabojarska:
> I wonder if we should bring Pkgfile headers under kres control 🤔

Sounds like a good idea, but the Pkgfile would need to be partially updated, not generated by Kres, right? Unless we move the whole thing to a template, which kres would use to generate the final Pkgfile :thinking:

frezbo:
> > I wonder if we should bring Pkgfile headers under kres control 🤔
> 
> Sounds like a good idea, but the Pkgfile would need to be partially updated, not generated by Kres, right? Unless we move the whole thing to a template, which kres would use to generate the final Pkgfile 🤔

yeh just partial is easier

PR #1620, opened 2026-07-17
feat: enable UFSHC and some other options

### LINKED ISSUE #1619: kernel: enable UFS host controller support (CONFIG_SCSI_UFSHCD)
## Problem

The Talos kernel currently has no UFS (Universal Flash Storage) support:

```
# CONFIG_SCSI_UFSHCD is not set
```

(`kernel/build/config-amd64`, same for arm64)

A growing number of low-cost x86 mini PCs ship with UFS as their only internal
storage. A concrete example is the Minisforum S100 (Intel N100 / Alder Lake-N,
64 GB UFS 2.1, 2.5GbE with PoE), which is otherwise a nice fit for small Talos
edge/homelab nodes. With current Talos images the internal disk is simply
invisible, so Talos cannot be installed on these machines at all.

Because the driver is not even built as a module, there is no way to work
around this with a system extension or Image Factory schematic — it requires a
kernel config change.

## Proposed change

Enable the UFS core and the PCI host controller glue (built-in, since this is
boot-critical root storage):

```
CONFIG_SCSI_UFSHCD=y
CONFIG_SCSI_UFSHCD_PCI=y
```

The mainline `ufshcd-pci` driver has supported the Intel Alder Lake UFS
controllers (8086:51FF / 8086:54FF, `ufs_intel_adl_hba_vops`) for years, so no
out-of-tree code is needed — it is purely a config enablement.

On arm64, UFS controllers are typically platform devices rather than PCI; I'd
suggest enabling at least `CONFIG_SCSI_UFSHCD=y` there as well, with platform
glue (e.g. `CONFIG_SCSI_UFSHCD_PLATFORM`) as a follow-up if there is demand
from SBC users.

## Notes

- UFS devices are exposed through the SCSI stack as regular `/dev/sd*` disks,
  so no changes on the Talos side should be needed beyond the kernel config.
- I'm happy to submit a PR for this, and I can test on real hardware
  (Minisforum S100).

smira:
Fixes #1619

```
kernel/drivers/md/dm-service-time.ko
kernel/drivers/target/target_core_mod.ko
kernel/drivers/ufs/core/ufshcd-core.ko
kernel/drivers/ufs/host/ufshcd-pci.ko
kernel/drivers/vhost/vhost_scsi.ko
```

Plus, for amd64:

```
kernel/drivers/devfreq/governor_simpleondemand.ko
```


dsseng (approved):

kernel/build/config-amd64
```diff
@@ -6383,7 +6405,21 @@ CONFIG_RPMSG_VIRTIO=y
 # end of Qualcomm PM Domains
 # end of PM Domains
 
-# CONFIG_PM_DEVFREQ is not set
+CONFIG_PM_DEVFREQ=y
+
+#
+# DEVFREQ Governors
+#
+CONFIG_DEVFREQ_GOV_SIMPLE_ONDEMAND=m
```
dsseng, line 6413: Why do we enable this only this on amd64? arm64 enables all the devfreq governors and I guess if there're consumer drivers it'd be interesting to be able to enable any of those

I don't know if any PC-compatible hardware uses devfreq though

smira: I don't know either, it was auto-selected, so I stayed conservative for now. I the need arises, we can add more of these

frezbo (approved):

Issue #116, opened 2020-05-06
Cilium now requires CONFIG_NETFILTER_XT_MATCH_SOCKET


Ulexus:
As per their [instructions](https://docs.cilium.io/en/v1.7/install/system_requirements/#linux-kernel), Cilium now requires the XT socket match capability for fast Layer7 service.

PR #1625, opened 2026-07-23
feat: enable CONFIG_NFT_SOCKET in the kernel

### LINKED ISSUE #1624: Enable CONFIG_NFT_SOCKET (nftables socket match) - required by ovn-kubernetes UDN host isolation
**What**

`kernel/build/config-amd64` (and `config-arm64`) currently has:

```
# CONFIG_NFT_SOCKET is not set
```

Please enable it (`CONFIG_NFT_SOCKET=y`, consistent with how the rest of netfilter is built in Talos).

**Why**

Without it, **every** nftables `socket` match fails to load - not just the cgroup form. That blocks
ovn-kubernetes' UserDefinedNetwork (UDN) host isolation, which programs:

```
add rule inet ovn-kubernetes udn-isolation \
  socket cgroupv2 level 2 <kubelet-cgroup> ip daddr @udn-pod-default-ips-v4 accept
```

On Talos this returns `Error: Could not process rule: No such file or directory`. There is no flag in
ovn-kubernetes to disable UDN host isolation, and the failure is fatal: it stops the node network
controller, so **every `ovnkube-node` pod crash-loops** and network segmentation is unusable. (The
crash surfaces as an unrelated nil-pointer panic in `metrics.MonitorIPSec`, which makes it hard to
find - worth mentioning so others searching land here.)

**The config looks internally inconsistent**

Talos already enables everything around it - only the nftables expression wrapper is off:

```
CONFIG_SOCK_CGROUP_DATA=y
CONFIG_NF_SOCKET_IPV4=y
CONFIG_NF_SOCKET_IPV6=y
CONFIG_NETFILTER_XT_MATCH_SOCKET=y     # legacy xtables equivalent, enabled
CONFIG_NFT_TPROXY=y
# CONFIG_NFT_SOCKET is not set         # <- this one
```

`NFT_TPROXY=y` with `NFT_SOCKET=n` is an unusual split; those are normally enabled together, since
transparent-proxy rulesets typically need both. `NETFILTER_XT_MATCH_SOCKET` (the iptables-era
equivalent) is already on - this is the nftables-era counterpart.

**Precedent**

`CONFIG_NETFILTER_XT_MATCH_SOCKET=y` was added for the same class of reason (a CNI required it) in
siderolabs/pkgs#116 -> siderolabs/pkgs#117, merged the same day. This is the nftables successor to
that option.

**Not Talos-specific in cause, but Talos is affected**

Every mainstream distro ships it as a module: Ubuntu 22.04/24.04, Debian, Fedora, RHEL 9 /
CentOS Stream 9 all set `CONFIG_NFT_SOCKET=m`. Flatcar has the same gap and an open RFE for exactly
this ovn-kubernetes case (flatcar/Flatcar#2172). Azure Linux enabled it on request
(microsoft/azurelinux#16869).

**Reproducer** (any Talos node, privileged pod, `nft` v1.1.6):

```
# nft add table inet t; nft add chain inet t c
# nft add rule inet t c accept                          -> rc=0   (nftables itself is fine)
# nft add rule inet t c socket mark 1 accept            -> rc=1   Error: No such file or directory
# nft add rule inet t c socket cgroupv2 level 1 kubepods accept
                                                        -> rc=1   Error: No such file or directory
# zcat /proc/config.gz | grep NFT_SOCKET
# CONFIG_NFT_SOCKET is not set
```

The same nft binary on Ubuntu 6.8.0-107-generic installs all three rules successfully. Note the
failure is not a cgroup-path problem: `socket mark 1`, which has no cgroup component, fails
identically, and a genuinely bad cgroup path produces a *different* error
(`cgroupv2 path fails: No such file or directory`).

Measured on Talos v1.12.7, kernel 6.18.24-talos.

imusmanmalik:
## What

Enable `CONFIG_NFT_SOCKET=y` in `kernel/build/config-amd64` and
`kernel/build/config-arm64`.

## Why

The nftables socket match is currently off:

    # CONFIG_NFT_SOCKET is not set

while every neighbouring netfilter match is on:

    CONFIG_NF_SOCKET_IPV4=y
    CONFIG_NF_SOCKET_IPV6=y
    CONFIG_NETFILTER_XT_MATCH_SOCKET=y   # legacy iptables equivalent
    CONFIG_NFT_TPROXY=y                  # normally paired with NFT_SOCKET

NFT_SOCKET is the nftables successor to the already-enabled
NETFILTER_XT_MATCH_SOCKET, and it is normally enabled alongside
NFT_TPROXY. Turning it on removes an inconsistent split.

Without it, all nftables `socket` matches fail to load. This breaks
ovn-kubernetes UserDefinedNetwork host isolation, which programs:

    socket cgroupv2 level 2 <kubelet-cgroup> ip daddr @udn-pod-ips accept

so kubelet health probes can reach primary-UDN pods. On Talos this
returns "Error: Could not process rule: No such file or directory".

## Reproducer (Talos v1.12.7, kernel 6.18.24-talos, nft v1.1.6)

    # nft add table inet t; nft add chain inet t c
    # nft add rule inet t c accept                       -> rc=0
    # nft add rule inet t c socket mark 1 accept         -> rc=1  No such file or directory
    # nft add rule inet t c socket cgroupv2 level 1 kubepods accept
                                                        -> rc=1  No such file or directory
    # zcat /proc/config.gz | grep NFT_SOCKET
    # CONFIG_NFT_SOCKET is not set

`socket mark 1` fails with no cgroup component, so this is the socket
match itself, not a cgroup-path problem.

The absence of `nft_socket.ko` is not the signal - `CONFIG_NFT_TPROXY=y`
is built in and has no `nft_tproxy.ko` either, yet works.

## Precedent

`CONFIG_NETFILTER_XT_MATCH_SOCKET=y` was added for a similar CNI need
(#116, #117). Ubuntu, Debian, Fedora, RHEL 9 and CentOS Stream 9 ship
NFT_SOCKET as a module; Azure Linux enabled it on request
(microsoft/azurelinux#16869); Flatcar has the same gap
(flatcar/Flatcar#2172). `=y` (not `=m`) matches the existing
`CONFIG_NFT_TPROXY=y` and avoids a module-load dependency.

Closes #1624


smira (approved):
thank you!

PR #1634, opened 2026-08-04
fix: iptables with Cilium

smira:
See https://github.com/siderolabs/talos/issues/13935

This partially reverts the change fbb73601, it turns out we only had to drop the `_LEGACY` config options to disable the iptables interfaces, while `iptables_nft` still uses some `xtables` modules which were also disabled as part of the change.

On Talos side, I would like to add a regression test for those once we bring the new kernel in.


shanduur (approved):

frezbo (approved):

smira:
Verified this with Talos integration test, which I will PR separately.

PR #1635, opened 2026-08-04
feat: support confiuring gnu mirror urls

frezbo:
GNU official mirrors are mostly down, bring in support to make it configurable and use kernel.org gnu mirror.


.kres.yaml
```diff
@@ -83,6 +83,12 @@ spec:
       - nonfree-kmod-nvidia-lts-pkg
       - nonfree-kmod-nvidia-production-pkg
   reproducibleTargetName: reproducibility
+  extraBuildArgs:
+    - GNU_MIRROR_URL
+  makefile:
+    extraVariables:
+      - name: GNU_MIRROR_URL
+        defaultValue: https://storm.goose-piano.ts.net:5008
```
smira: should we use mirrors.kernel.org?

frezbo: 🤣 I used my local

smira: I think we should use mirros.kernel.org if possible... unless we want to promote yours to be production

smira (approved):

PR #1636, opened 2026-08-04
feat: enable CONFIG_FSCACHE and CONFIG_NFS_FSCACHE

ChristianJacobsen:
Enables NFS fscache. The userspace daemon ships as a Talos system extension.


smira (approved):

ChristianJacobsen:
@smira, I wonder if the build timeouts might be explained by last (work) week's Github outages... Would you mind triggering another run? It doesn't look like I can do it from my end.

PR #1638, opened 2026-08-06
feat: kernel: enable CONFIG_FS_ENCRYPTION

donch:
This PR enables CONFIG_FS_ENCRYPTION in the kernel config, needed for ceph-csi to support fscrypt-based encryption on CephFS volumes, currently unavailable without a fully custom kernel build.


smira (approved):
thank you!

PR #1651, opened 2026-08-16
feat: enable Bluetooth USB kernel modules

staners2:
## What

Enable the minimal modular kernel support required for Bluetooth LE over USB on
amd64 and arm64:

- the Bluetooth core and LE support;
- rfkill;
- the USB HCI transport;
- Intel, Realtek, and MediaTek vendor helpers.

The existing upstream USB autosuspend default remains enabled. The modules are
only loaded when an image includes and uses a corresponding system extension.

## Why

A Talos system extension can package kernel modules, but it cannot provide
modules that the matching Talos kernel build did not produce. USB passthrough
alone therefore exposes a controller to a workload without creating a Linux
HCI device.

The first consumer is siderolabs/extensions#1203, which packages only these
version-matched modules and their dependency metadata. Firmware and BlueZ are
kept outside this kernel change.

## Validation

- `BUILDX_BUILDER=talos-bt-buildkit gmake kernel-olddefconfig` completed for
  amd64 and arm64 and reported no changes to either generated configuration.
- A full multi-architecture kernel and extension build was previously completed
  against Linux `6.18.44-talos`; module signatures and `vermagic` matched that
  kernel.
- Runtime validation on amd64 with an integrated MediaTek MT7922 produced
  `hci0`, exposed the BlueZ adapter API, and completed an active BLE scan.

Realtek runtime validation is intentionally not claimed here.

## Security

This change adds opt-in modules only. It does not start BlueZ or D-Bus, grant
container privileges, or alter the Talos management API.


smira:
I would really not like the idea of adding Bluetooth support to Talos. Talos Linux is a server OS, and Bluetooth stack is complicated and full of CVEs, so for the sake of security I would really not have it even it means losing some niche usecases.

staners2:
Thank you for the feedback. I understand the concern about expanding the
kernel attack surface and the long-term CVE maintenance burden.

Before closing this direction, could you clarify whether the maintainers are
categorically against any Bluetooth support in the Talos kernel, or whether an
architecture with a strictly opt-in runtime boundary could be acceptable?

The intended boundary would be:

- Bluetooth core and USB HCI built only as modules, never built in;
- modules absent from the standard installer and delivered only through an
  explicitly selected `contrib` system extension;
- no module autoloading, BlueZ, D-Bus policy, or Bluetooth userspace in Talos;
- the smallest LE/USB driver set, with vendor firmware kept in separate
  extensions;
- explicit community maintainer ownership for compatibility and security
  updates.

If this is still outside the acceptable scope for Talos, I will treat that as
a firm project boundary and close the dependent extension PRs. If there is an
acceptable design boundary, what additional isolation or maintenance
requirements would you want before reconsidering it?

smira:
Adding things is easy, supporting it in the long run and maintaining it is not, so I'm not sure I fully understand how many users would benefit from having BT support. And if we do BT support, it should be full one, and I guess just kernel module is not enough? (I might be wrong on this one).

PR #1652, opened 2026-08-16
feat: kernel: enable MPTCP, sched_ext

dsseng:
Opt-in features enabling userspace to make use of new kernel features


kernel/build/config-arm64
```diff
@@ -1398,7 +1398,9 @@ CONFIG_IPV6_SEG6_BPF=y
 # CONFIG_IPV6_RPL_LWTUNNEL is not set
 # CONFIG_IPV6_IOAM6_LWTUNNEL is not set
 CONFIG_NETLABEL=y
-# CONFIG_MPTCP is not set
+CONFIG_MPTCP=y
```
smira, line 1401: I have a bad memory, isn't it the one we disabled previously?

I guess not, it's multipath TCP?

shanduur (approved):

kernel/build/config-arm64
dsseng: It is multipath TCP. Not sure, I'll check if we disabled it at some point

dsseng: Seems to have been off forever (5.6 introduced MPTCP support)

https://github.com/siderolabs/pkgs/blame/c9173bb3fa60ae23ad8f2c6e8cee0499a8c45077/kernel/build/config-amd64#L1414

smira (approved):

PR #1654, opened 2026-08-17
feat(kernel): enable CONFIG_MLX5_VFIO_PCI

maxpain:
## What

Flips `CONFIG_MLX5_VFIO_PCI` from unset to `m` in `config-amd64` and `config-arm64`.

## Why

`mlx5-vfio-pci` is the mlx5 VFIO *variant* driver. It passes a ConnectX VF
through exactly like `vfio-pci` does, but additionally implements the VFIO
migration v2 uAPI — device state save/restore, `PRE_COPY`/`RUNNING_P2P` states
and DMA dirty-page logging offloaded to the NIC.

With plain `vfio-pci` the VF has no migration region, so a guest holding an
SR-IOV VF cannot be live-migrated at all. That makes SR-IOV networking and
live migration mutually exclusive on Talos today, and the only workaround is a
fully custom kernel build.

Observed on Talos v1.13.4 (`6.18.34-talos`): the firmware side is already
willing (`VF_MIGRATION_MODE = MIGRATION_ENABLED`), but
`/lib/modules/6.18.34-talos/kernel/drivers/vfio/pci/` ships only `vfio-pci.ko`
and `vfio-pci-core.ko`, so the VF can only be bound to `vfio-pci`.

## Cost

The option is a small tristate building a single module,
`drivers/vfio/pci/mlx5/mlx5-vfio-pci.ko`. It `depends on MLX5_CORE` and
`select`s `VFIO_PCI_CORE`, both already `=m` in these configs, so no other
symbol changes and nothing is pulled into the base image — the module is only
loaded if an operator asks for it via `machine.kernel.modules`.

Both architectures are changed to keep the two configs in step.


smira (approved):
thank you!

PR #1526, opened 2026-04-28
kernel: macb silent TX stall candidate fixes (RFC patches from netdev)

lukaszraczylo:
## net: macb: candidate fixes for silent TX stall on BCM2712/RP1

Mirrors the netdev RFC `[RFC PATCH net-next 0/3] net: macb: candidate fixes for silent TX stall on BCM2712/RP1` ([lore](https://lore.kernel.org/netdev/cover.1777064117.git.lukasz@raczylo.com/T/)) byte-for-byte. Drops three patches into `kernel/build/patches/` against the existing `linux_version: 6.18.24` pin — apply cleanly, no source-tree changes.

Per the discussion on siderolabs/sbc-raspberrypi#91, scope is limited to **macb driver patches that have been submitted upstream**:

- `0001-net-macb-flush-PCIe-posted-write-after-TSTART-doorbe.patch` — flushes PCIe posted writes after the `TSTART` doorbell so the write reaches the MAC before the function returns.
- `0002-net-macb-re-check-ISR-after-IER-re-enable-in-macb_tx.patch` — re-reads ISR after `IER` re-enable in `macb_tx_poll()` to catch a `TCOMP` raised inside the IDR/IER mask window.
- `0003-net-macb-add-TX-stall-watchdog-as-defence-in-depth-s.patch` — per-queue `delayed_work` safety net that calls `macb_tx_restart()` if `tx_tail` hasn't advanced for ≥ 1 s while the ring is non-empty.

EEE PHY patches from raspberrypi/linux PR #7270 are **not** included — they are vendor-fork-only and target a dead-case bug that doesn't exist in mainline.

### Verification

- Apply test: clean `patch -p1` against `gregkh/linux@v6.18.24` (the tarball `Pkgfile` already pulls). All three hunks land first try.
- Build test: `make kernel PLATFORM=linux/arm64` against this PR completes end-to-end. Kernel + modules + signing + depmod + SPDX manifest all produced. drivers/net/ethernet/built-in.a (where macb_main.c lives) compiles with no new warnings.

### Runtime data

Running on a 24-node Raspberry Pi 5 fleet (Talos `v1.12.7-k-6.18.24-macb`, kernel `6.18.24-talos`):

- Series deployed 2026-04-24 19:00 UTC. ~1,900 cumulative node-hours of clean telemetry as of opening this PR.
- **Zero** silent TX wedges (the failure mode the series targets; pre-patch each event required a hardware power-cycle).
[...]

PR #1655, opened 2026-08-17
feat(kernel): add raspberry pi rp1 pwm driver

geoffdavis:
## What? (description)

Adds the `pwm-rp1` driver to the kernel so the Raspberry Pi 5's PWM controller
(the RP1 southbridge block, `compatible = "raspberrypi,rp1-pwm"`) has a driver,
and enables `CONFIG_PWM_RP1=y` on arm64. The driver is taken verbatim from the
Raspberry Pi downstream kernel (`rpi-6.18.y`, `drivers/pwm/pwm-rp1.c`, © 2023
Raspberry Pi Ltd, GPL-2.0); the Kconfig/Makefile wiring is placed to match
mainline.

## Why? (reasoning)

On the Raspberry Pi 5 the Active Cooler fan is driven through the RP1's PWM
block. The firmware DTB already ships an **enabled** `cooling_fan` node that
references `rp1 pwm@9c000`, and the kernel already has the consumer side
(`CONFIG_SENSORS_PWM_FAN=y`). What is missing is a **provider**: mainline Linux
has no RP1 PWM driver (checked through v6.19 — `drivers/pwm/Kconfig` has no RP1
entry), it exists only downstream.

The result on a stock `rpi_5` Talos node:

- `/sys/kernel/debug/devices_deferred` → `cooling_fan  pwm-fan: Could not get PWM`
- `/sys/class/pwm/` empty, no `cooling_device`, no fan telemetry
- the thermal governor has **no control of the fan** — nodes idle at 68–75 °C
  and peak at 86.5 °C, past the SoC soft-throttle threshold, regardless of load

This is the same class of Raspberry Pi 5 RP1 gap already addressed in this repo
by the macb EEE backport (`fix: backport macb IEEE 802.3az EEE support for
Raspberry Pi 5 RP1`), and the fix is analogous: carry the downstream driver.

## Verification

Built for arm64 and deployed to `rpi_5` nodes. With the driver present:

| check | before | after |
| --- | --- | --- |
| `/sys/class/pwm/` | empty | `pwmchip0` |
| `cooling_device0` | absent | present, `type=pwm-fan` |
| `devices_deferred` | `pwm-fan: Could not get PWM` | empty |
| fan hwmon | none | `pwmfan` |
| governor control | none | active (fan state tracks SoC temp) |

Node idle temperature dropped below its previous floor once the governor could
drive the fan.

## Acceptance

- [x] you formatted your code (`make fmt`) — n/a, patch + kconfig only
- [x] you linted your code (`make lint`)
- [ ] you ran conformance (`make conformance`)
- [x] driver is GPL-2.0, unmodified from the downstream source; only Kconfig/
      Makefile placement adjusted for mainline

## Conformance note

`make conformance` passes every check except `GPG Identity`, which fails
because I am not a member of the `siderolabs` org (the commit is DCO
signed-off and its signature verifies). This is expected for an external
contributor — cf. #1526, another Raspberry Pi 5 RP1 kernel patch that merged
with the same `GPG Identity` result.


smira:
Hi, thanks for your PR.

If the driver is not the Linux mainline kernel, we would not accept it most likely, as we have no resources to maintain 3rd party patches.

If it got mainlined e.g. in 7.2, we could accept this patch.

barnes-c:
Hi @smira,
I wanted to make a similar PR and saw that there is an [upstream series](https://patchwork.ozlabs.org/project/linux-pwm/list/?series=514600) on its way so I held off:

* [[v7,1/3]](https://patchwork.ozlabs.org/project/linux-pwm/patch/67d51a7210f0d7410dbe34e3d93813db1fecd622.1784539788.git.andrea.porta@suse.com/) dt-bindings: pwm: Add Raspberry Pi RP1 PWM controller
* [[v7,2/3]](https://patchwork.ozlabs.org/project/linux-pwm/patch/a83d65b2d91865eb722337c547c94dce175e6086.1784539788.git.andrea.porta@suse.com/) pwm: rp1: Add RP1 PWM controller driver
* [[v7,3/3]](https://patchwork.ozlabs.org/project/linux-pwm/patch/895152a8a7b541279519392b5e7983a861a7bd13.1784539788.git.andrea.porta@suse.com/) arm64: dts: broadcom: rpi-5: Add RP1 PWM node <-- not needed, the DT is already in sbc-raspberrypi

It from SUSE, who are upstreaming the rest of RP1 support

The code is also quite different from the downstream driver proposed here, e.g. different Kconfig symbol (`PWM_RASPBERRYPI_RP1` vs `PWM_RP1`) and different deps (`MISC_RP1` + `REGMAP_MMIO` vs `ARCH_BCM2835`).

I already tried carrying [1/3]+[2/3] here. They apply cleanly to `6.18` but don't compile: the driver uses `HZ_PER_GHZ`, added by
[5083dba0fde5](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=5083dba0fde5446c00ee1a82a3911c8f88a2c72e) ("units: Add `HZ_PER_GHZ`"), which first appears in `7.0` and is absent from `6.18/6.19`. So the upstream series would need an extra backport for now.

Would you consider taking the upstream patches once they appear in linux-next or an -rc or is that too early?

smira:
> Would you consider taking the upstream patches once they appear in linux-next or an -rc or is that too early?

As long as it lands in the rc/next, it could be backported. But given the timeline the next Talos release (1.15) will be based on 7.<latest>, so it will happen on its own.

geoffdavis:
Thanks @smira, @barnes-c — that all makes sense, closing this. Agreed the SUSE mainline series ([`pwm: rp1` v7](https://patchwork.ozlabs.org/project/linux-pwm/list/?series=514600)) is the right path, and since Talos 1.15 will be on 7.x it'll come in for free — no downstream patch to maintain. I'll carry this locally in the interim. Appreciate the pointer to the upstream work.

PR #1663, opened 2026-08-23
feat: enable CONFIG_XFRM_INTERFACE in the kernel

imusmanmalik:
Two lines, one per arch:

```
-# CONFIG_XFRM_INTERFACE is not set
+CONFIG_XFRM_INTERFACE=y
```

## Why

XFRM interfaces are the only part of the XFRM stack still switched off. `XFRM`,
`XFRM_USER`, `XFRM_ALGO`, `XFRM_OFFLOAD` and `XFRM_STATISTICS` are all enabled, so this
just fills the remaining gap for route-based IPsec.

Without it you can only do policy-based IPsec, where the kernel picks the tunnel from the
policy selectors. That works for a single tunnel. It does not work for a redundant pair,
because the policy database holds one entry per selector pair - the same subnets to a
second peer are refused:

```
cannot install kernel policy fd00:42::/56===fd00:31::/64
in use by established Child SA with routing routed-tunnel
```

The standby then waits on whatever retry timer its userspace daemon uses instead of on the
routing layer, so failover takes tens of seconds and there is no ECMP. With an XFRM
interface the tunnel is a netdev, routing decides what goes into it, and a routing protocol
with BFD can run over it, which is how you get both back.

## Where I hit it

Talos v1.13.8 with libreswan 5.3, twelve nodes, node-to-node IPsec plus a redundant pair of
egress tunnels. libreswan is built with XFRMI support - it says so in its own version
string - and rejects the config at load because the kernel cannot provide the interface:

```
Starting Pluto (Libreswan Version 5.3 IKEv2 IKEv1 XFRM XFRMI esp-hw-offload ...)
"tuntest": failed to add connection: ipsec-interface=86 is not supported
```

Any IPsec implementation that wants route-based tunnels lands in the same place - strongSwan
uses the same kernel feature for its `xfrm_interface` option. Nothing about the change is
specific to one daemon.

## Risk

It is a software netdev with no hardware dependency, and depends only on `XFRM` and `IPV6`,
both already enabled. Nothing changes unless something asks for an `ipsecN` interface.
`CONFIG_XFRM_STATISTICS` was enabled the same way in #1052.


smira (approved):
I changed it to be a module, as it's an exotic option

PR #1668, opened 2026-08-26
feat(kernel): enable Apple Silicon platform drivers

eseiker:
## Summary

- enable the mainline Apple Silicon platform in the generic arm64 kernel config
- build the platform, interrupt, IOMMU, PCIe, NVMe, power-management, SMC, and serial-console drivers into the kernel
- retain Talos' existing 4 KiB arm64 page size

## Motivation

The earlier Apple Silicon platform request, siderolabs/talos#10229, was closed as not planned in January 2025. At the time, the key blocker called out in that discussion was that the upstream kernel could not boot Apple Silicon and was not close to doing so.

That technical situation has materially changed. By Linux 6.18, the Apple AIC, DART, PCIe, NVMe, SART/RTKit, PMGR, and MacSMC support needed by the tested Mac14,13 (2023 Mac Studio, M2 Max) is present upstream. A Talos v1.13.9 image using Linux 6.18.44 with no downstream Apple kernel patches now boots the machine and reaches a working Talos API, kubelet, and Kubernetes node. The [Asahi M2 feature matrix](https://github.com/AsahiLinux/docs/blob/main/docs/platform/feature-support/m2.md) also records upstream Linux 6.18 device-tree and Ethernet support for the 2023 Mac mini and Mac Studio families.

This PR does not attempt to reopen the broader question of official support for every Apple model. It is a config-only step that makes the Apple drivers already available in the upstream kernel usable by the generic Talos arm64 build.

Keeping this as a kernel configuration change avoids downstream kernel patches and does not introduce a separate page-size variant. Boot-critical drivers are built in because interrupt, IOMMU, PCIe, and NVMe support are needed before a root filesystem or modules can be loaded.

## Why this matters

Together with the existing Asahi/m1n1/U-Boot boot chain, this removes the kernel-config blocker to using compact, power-efficient Mac mini and Mac Studio systems as bare-metal Talos Kubernetes nodes. That makes clusters built from these machines practical without nesting Talos in virtual machines or maintaining a downstream Asahi kernel fork, while retaining Talos' immutable and declarative lifecycle.

Hardware enablement still varies by model, so this is not a blanket compatibility claim. The hardware validation for this change is specifically Mac14,13 (2023 Mac Studio, M2 Max); other Mac mini and Mac Studio models should be validated independently.

## Validation

- ran `gmake kernel-olddefconfig PLATFORM=linux/arm64 USERNAME=eseiker CI_RELEASE_TAG=` and committed the normalized config
- ran `gmake check-dirty`
- booted Talos v1.13.9 with a mainline Linux 6.18.44 kernel using the same Apple driver set on a Mac14,13 (2023 Mac Studio, M2 Max)
- verified the internal Apple NVMe device and Apple PCIe-attached Aquantia/Atlantic Ethernet
- verified Talos API, kubelet, and Kubernetes node operation
- verified a 4 KiB kernel/userspace page size with `getconf PAGESIZE` returning `4096`
- observed all probed Apple DART instances initialize with `pagesize 4000`, `bypass support: 1`, and `bypass forced: 1`
- you can observe the prebuilt images here https://github.com/eseiker/talos-asahi

## Scope and limitations

This is kernel configuration enablement, not a claim of official Talos support for every Apple Silicon model. The existing `DRM_SIMPLEDRM` and framebuffer console provide firmware-framebuffer boot diagnostics. Apple ADP/backlight, audio, touch, and other desktop-oriented drivers remain disabled; the mainline ADP driver targets pre-DCP controllers rather than the Mac14,13 (2023 Mac Studio, M2 Max) main display.

Apple DART uses identity domains with a 4 KiB kernel because the hardware IOMMU granule is 16 KiB (`4000` in the kernel log is hexadecimal). As a result, DART-backed devices do not receive DMA isolation in this mode. That is the existing mainline fallback which makes the generic 4 KiB arm64 kernel usable on this hardware; this change does not alter the DART implementation.


smira:
@eseiker I've applied some changes/fixes to your PR

Also I believe this won't boot on Apple Silicon due to missing UEFI support, so it needs DTB/U-Boot?

eseiker:
That's where I leveraged upstream [asahi-installer](https://github.com/AsahiLinux/asahi-installer). Asahi installer's [UEFI only install consists U-boot and DTBs](https://github.com/eseiker/talos-asahi#2-install-the-official-asahi-uefi-environment), and some handlings with META/STATE partition to assign machine UUID and GPT alignment by prepare.efi in [my repository](https://github.com/eseiker/talos-asahi), now it can boot.

smira:
This will probably stay 1.15+ only, and it should land 7.x kernel, so hopefully by 1.15.0 the support will be even better with upstream Linux updates.

smira (approved):

dsseng (approved):
7.x kernels are also going to have HWMON (sensors) support - I would suggest keeping this in mind to enable and test it once Talos kernel version is updated

PR #1682, opened 2026-09-06
feat: enable earth-pt3 dvb driver

5ym:
## What

Enable `CONFIG_DVB_PT3=m` for amd64.

## Why

Earthsoft PT1/PT2/PT3 are ISDB-T/S PCIe tuner cards, widely used for terrestrial and satellite TV recording in Japan. Talos currently cannot drive them because the in-tree `earth-pt3` driver is not enabled, so there is no way to run a recording workload on Talos with this hardware.

Everything the driver needs is already built as a module in this config, so the change is a single line:

| symbol | before | after |
| --- | --- | --- |
| `CONFIG_DVB_PT3` | not set | `m` |
| `CONFIG_DVB_TC90522` (demodulator) | `m` | unchanged |
| `CONFIG_MEDIA_TUNER_QM1D1C0042` (ISDB-S tuner) | `m` | unchanged |
| `CONFIG_MEDIA_TUNER_MXL301RF` (ISDB-T tuner) | `m` | unchanged |

Only `config-amd64` is touched: the card is PCIe and there is no practical arm64 use.

## Follow-up

A `dvb/pt3` system extension that ships `earth-pt3.ko` and its dependencies, modelled on the existing `dvb-cx23885` extension, needs this to land first. I will open that PR against `siderolabs/extensions` once the kernel builds the module.

I do not have a build environment large enough to compile the kernel locally, so this relies on CI.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01JgGYf5qbjDXYhtzcpLsGvv


smira (approved):
👍

PR #1692, opened 2026-09-17
feat(kernel): enable CONFIG_NO_HZ_FULL

maxpain:
Switches the tick mode from `NO_HZ_IDLE` to `NO_HZ_FULL` on amd64 and arm64. It selects `RCU_NOCB_CPU` and `VIRT_CPU_ACCOUNTING_GEN`; the rest of the diff is `make kernel-olddefconfig`.

Without it `nohz_full=` is dropped at boot ("Housekeeping: nohz unsupported. Build with CONFIG_NO_HZ_FULL") and `rcu_nocbs=` is unknown, so isolated cores running DPDK/SPDK pollers or pinned vCPUs still take the 250 Hz tick. The only workaround is a custom kernel.

No behavior change unless `nohz_full=` or `rcu_nocbs=` is passed. The Kconfig help says: "without passing the nohz_full parameter, this behaves just like NO_HZ_IDLE. If you're a distro say Y". Fedora, RHEL, Debian, Arch and openSUSE ship it enabled.

Hardening check passes; not boot-tested.


smira:
RCU_NOCB_CPU=y is pulled in unconditionally, so "no behavior change unless nohz_full=/rcu_nocbs= is passed" is slightly too strong. With nothing offloaded, call_rcu() now goes through rcu_nocb_try_bypass() and the locking helpers become runtime rcu_rdp_is_offloaded(rdp) branches where they were compile-time no-ops. RCU_NOCB_CPU_DEFAULT_ALL=n is correctly set so no kthreads spawn and the cost is a few predictable branches — same trade-off Fedora/RHEL/Debian accept.

smira (commented):
👍

smira (approved):

